PR #3, opened 2019-10-22
feat(editor): add monaco editor

JeanBarriere:
- add editor view using monaco-editor component
- add monaco-editor component based from single monaco instance in the monaco plugin
- add monaco plugin instantiating the storyscript language as the unique language loaded
- add types
- add unit testing

____
_Solving Jira issue **PL-1022**_

When I land on the playground,
I am presented with an example read-only story,
So that I can quickly understand the syntax of the language, and grasp its scope

### Acceptance Criteria
**Given** I navigate to the playground
**Then** I am presented with the following example story:

```
redis set key: "count" value: 0
when http server listen path: "/count" as req
  count = (redis get key: "count").result + 1
  redis set key: "count" value: count
  req write content: "This page has been loaded {count} times"
```
___
Editor should exist as in the design document with syntax highlighting


williammartin:
### Acceptance

### Acceptance Criteria
**Given** I navigate to the playground
**Then** I am presented with the following example story:

```
redis set key: "count" value: 0
when http server listen path: "/count" as req
  count = (redis get key: "count").result + 1
  redis set key: "count" value: count
  req write content: "This page has been loaded {count} times"
```

image

LGTM

PR #7, opened 2019-10-24
[PL-1115] - Prospective Storyscripter does not think the logs are a terminal

Arinono:
When I click the Deploy button,
I do not see `story deploy` terminal prompt in the logs,
So I don’t think I’m looking at a terminal

## Acceptance

**When** I click the Deploy button
**Then** I see the logs appear incrementally, but with the first line removed:

```
$> story deploy
```

### DIFF
--- a/src/models/StorySample.ts
+++ b/src/models/StorySample.ts
@@ -3,7 +3,6 @@ export interface IStorySample {
 }
 
 export interface IStoryLogs {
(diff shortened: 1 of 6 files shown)


williammartin:
### Acceptance

**When** I click the Deploy button
**Then** I see the logs appear incrementally, but with the first line removed:

image

LGTM

PR #21, opened 2019-10-28
chore(sentry): add Sentry DSN to env variable config

matthewhudson:
@JeanBarriere Here's the link to the [`playground`](https://sentry.io/asyncy/playground/getting-started/javascript-vue/) project in Sentry.


Arinono:
We should probably not push in plain text (and regenerate ?) this key and just put it in Netlify environment variables 🌝

.env
```diff
@@ -1 +1,2 @@
-VUE_APP_SSL_ENABLE=true
\ No newline at end of file
+VUE_APP_SSL_ENABLE=true
+SENTRY_DSN=https://f20291174ff6475cab01fe6afde7b749@sentry.io/1799311
```
JeanBarriere: we may want to add it in the netlify build instead of here, wdyt ?

JeanBarriere:
Also, can you change this variable to VUE_APP_SENTRY_DSN ? I think it won't go there otherwise (and I'm thinking it may be the reason we don't have sentry going up on the hub)

matthewhudson:
> Also, can you change this variable to VUE_APP_SENTRY_DSN ? I think it won't go there otherwise (and I'm thinking it may be the reason we don't have sentry going up on the hub)

@JeanBarriere 👍

RE: AFAICT the [Sentry DSN isn't a secret](https://forum.sentry.io/t/dsn-private-public/6297).

JeanBarriere:
LGTM, just missing the docker build env

PR #39, opened 2019-11-04
PL-1113 - Prospective storyscripter gets the story autotyped

JeanBarriere:
- [x] add typing method
- [x] add unit testing
- [x] add sample module in store
- [x] write if sample name is not stored
- [ ] add e2e test

___
I see the script autotype incrementally,
So I can parse the script bit by bit, rather than as a large block

### Acceptance
**Given** I open an example
**Then** I see the script incrementally start to appear as if being typed
**And** When I navigate away and return to the page
**Then** I don’t see it being typed again

___

**Note**
@williammartin should this be persistant ? I understand "navigate away" as going to the welcome page and going back to the playground. Is that the case ? Or by "navigate away" you mean closing and reopening the playground ?

### DIFF
--- a/src/store/modules/Samples.ts
+++ b/src/store/modules/Samples.ts
@@ -1,0 +1,16 @@
+const state = {
+  written: []
+}
+
+const getters = {
+  getSamplesWritten: (state: any) => state.written,
+  isSampleWritten: (state: any) => (sample: string) => state.written.includes(sample)
+}
+
+const mutations = {
+  writeSample: (state: any, sample: string) => !state.written.includes(sample) && state.written.push(sample)
+}
+
+export default {
+  state, getters, mutations
+}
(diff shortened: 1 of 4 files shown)


williammartin:
### Acceptance

**Given** I open an example
**Then** I see the script incrementally start to appear as if being typed
**And When** I navigate away and return to the page
**Then** I don’t see it being typed again

^ This worked as I saw the script being typed behind the tips (or with `skipIntro=true` writing directly)

I think it gave us a few good ideas moving forward for after GHU when the tips become more relevant.

Here's the rough outline of how I see this integrating:
 - Arrive on the page with blank syntax editor and first tool tip (welcome to Storyscript...) is visible
 - I close the first tip and the first line starts writing:

```
redis set key:"count" value:0
```

 - I then see the next tip appear relating to the redis (as is today)
 - I close that tip and the next block writes with tips appearing incrementally

Let's close this PR as it's done the job for exploration.

Cheers!

PR #43, opened 2019-11-05
Add ephermal file sharing example + raw-loader + CircleCI verification

wilzbach:


JeanBarriere:
@Arinono can you update the files story object to match your changes ?

Then we can merge this

PR #45, opened 2019-11-05
chore(logs): set font to monospace on logs

JeanBarriere:
- [x] change font from sans-serif to monospace on logs


williammartin:
### Acceptance

**When** I look at the logs from deploy
**Then** I see they are monospaced

image

LGTM. Have taken this to design as an open question so let's leave this PR hanging and we can merge later if design approve.

JeanBarriere:
from @aydaoz's feedback:
> lets keep as in design. only code is monospace. logs basically the activity window so no need code font

PR #54, opened 2019-11-06
deploy dev to master

JeanBarriere:
- [x] update version
- [x] create release and tag


williammartin:
**When** I navigate to `playground.storyscript.io`
**Then** I presented with the playground welcome page

**And When** I access it over `https` 
**Then** I get a valid cert

image

On `playground.storyscript.io` and have a valid certificate. LGTM

PR #56, opened 2019-11-06
feat(welcome): icons for services used in story should appear on card on welcome page

Arinono:


src/views/Welcome.vue
```diff
@@ -52,7 +52,21 @@
               </s-text>
             </div>
           </div>
-          <div class="w-full flex justify-end items-center p-3">
+          <div class="w-full flex justify-between items-center pl-8 pr-3 py-3">
```
JeanBarriere:
```suggestion
          <div class="w-full flex justify-between items-center p-3">
```

PR #59, opened 2019-11-06
feat(logs+archi): no logs more cards

Arinono:
Exploring a refactor around the logs and the architecture diagram


williammartin:
Not doing specific acceptance here because it was just for exploration, but looks great and have created some more features out of this. Feel free to close out. Thanks.

PR #61, opened 2019-11-07
feat(samples): add Todo-backend sample

wilzbach:
The app is fully functional (as is!) and available at https://electronic-tin-9248.storyscriptapp.com

The story will even work directly for users as it auto-bootstraps a mongodb db on the pod (like redis), s.t. users can immediately start playing (of course with the limitation that we don't provide any guarantees about the data + it can only run in a single pod).

Details: https://github.com/wilzbach/mongodb

Further note: the sample does _not_ pass the [Todo backend testsuite](https://www.todobackend.com/contribute.html) as this requires CORS and currently there's no other way than manually adding it to every block.


JeanBarriere:
We'll be adding more examples when we'll be implementing graphQL and pulling them from a database. For now, closing this PR and referencing it [here](https://www.pivotaltracker.com/story/show/169919867)

PR #69, opened 2019-11-11
Add gmail autolabel example

williammartin:
Draft for the gmail autolabel example.


src/samples/autolabel.ts
```diff
@@ -0,0 +1,20 @@
...
+  services: [
+    'gmail',
+    'machinebox',
+    'redis'
+  ],
+  stories: ['autolabel'],
+  code,
+  collaborators: defaultCollaborators,
+  description:
+    'Train a machine learning model to predict and apply labels to a gmail inbox using the gmail, machinebox, and redis services.',
```
matthewhudson: Should we capitalize the proper nouns? EG `gmail` x2, `machinebox`, and `redis`

Arinono: I guess we should write the same as the brands are 👍

williammartin: Yeh the only reason I didn't do this was because it wasn't in the counter example and I figured I'd just fix them all at once in a `Proper Noun` PR rather than mixing counter changes into this PR.

PR #74, opened 2019-11-11
feat(editor): add fullscreen behavior

Arinono:
- [x] add fullscreen button and behavior
- [x] catch `Esc` keypress to exit fullscreen
- [x] unit testing
- [x] e2e testing


src/views/Playground/index.vue
```diff
@@ -9,32 +9,62 @@
...
+          id="enter-fullscreen"
+          icon="fullscreen"
+          width="20"
+          height="20"
+          color="text-indigo-50"
+          class="hover:bg-gray-10 rounded-full p-1"
+          clickable
+          @click="fullscreen = true"
+        />
+        <s-icon
```
JeanBarriere: maybe one icon only is possible ?

Arinono: Done 👍

PR #76, opened 2019-11-11
Move first tip to right of demo label

williammartin:


Arinono:
Screenshot 2019-11-11 at 19 43 15

👍

PR #82, opened 2019-11-12
improvement(pwa): improve caching strategies for POST requests

Arinono:


src/views/Login.vue
```diff
@@ -193,7 +193,7 @@ export default class Login extends Vue {
     this.error = false
     fetch('/', {
       method: 'POST',
-      headers: { 'Content-Type': 'application/x-www-form-urlencoded' },
+      headers: { 'Content-Type': 'application/x-www-form-urlencoded', 'Cache-Control': 'no-cache' },
```
williammartin: Do we want to keep this?

public/service-worker.js
```diff
@@ -0,0 +1,38 @@
...
+ * See https://goo.gl/nhQhGp
+ *
+ * The rest of the code is auto-generated. Please don't update this file
+ * directly; instead, make changes to your Workbox build configuration
+ * and re-run your build process.
+ * See https://goo.gl/2aRDsh
+ */
+
+importScripts('https://storage.googleapis.com/workbox-cdn/releases/4.3.1/workbox-sw.js')
+importScripts(
```
williammartin: Where do we get this file name from? How do we know it is going to be the same?

src/views/Login.vue
Arinono: We can remove it 😉

public/service-worker.js
Arinono: 🤷‍♂

williammartin: Look, I'll accept this if it works, but I'm _very_ suspicious of this. It doesn't seem right at all.

JeanBarriere:
I'm also suspecting there is something wrong about how we manage these here. Will search more about this when I come back so I can actually give a proper review to this

PR #83, opened 2019-11-12
feat(editor): set sane defaults

williammartin:


src/views/Studio/index.vue
```diff
@@ -140,8 +140,12 @@ export default class Studio extends Vue {
   private options: any = {
     readOnly: true,
     minimap: { enabled: false },
-    fontSize: 16,
-    automaticLayout: true
+    fontSize: 12,
+    automaticLayout: true,
```
matthewhudson: 👍👍👍 `automaticLayout`

williammartin:
@Arinono What's up with the pending statuses here? Are they stuck?

Arinono:
@williammartin GitHub Actions seem to be blocked [here](https://github.com/storyscript/studio/actions?page=2). I canceled the last workflow, it was still running even if all the test suites were done

williammartin:
Can we merge?

Arinono:
Yes, this is safe changes 😉

PR #84, opened 2019-11-12
feat: add body-scroll-lock

Arinono:
cc @williammartin @matthewhudson


Arinono:
I added `body-scroll-lock` but the behavior is not good:
IMG_0148
Here the scroll is locked

IMG_0149
Here it is not 😞

JeanBarriere:
@Arinono what should we do with this ?

Arinono:
@JeanBarriere I need to do some more research to make it work in every use case. The lib I'm using is working 100% in their demo on my iPad. So I need to check why in our case it is not.

Arinono:
I was unable to make it work on iPad with iOS (iPadOS) 13. [This example](https://bodyscrolllock.now.sh/) (which works on Safari on macOS) does not work either. According to [this issue](https://github.com/willmcpo/body-scroll-lock/issues/134), and from what I was able to test (2 other packages + manually tweaking CSS/JS events) this is currently not possible to do.

I'm still leaving this PR ready for review and merge because it still fixes this behavior on Safari macOS, so it's a good take. And if the `body-scroll-lock` package gets updated, this will trigger a new PR from `renovate`.

PR #87, opened 2019-11-13
feat(samples): add autolabel comments

williammartin:


Arinono:
## Acceptance
**When** I look at the comments for the gmail example
**Then** I see the following exchange:

>Hey Inkie! Storyscript argument names are optional if the variable being provided has the same name as the function argument so I think where we have `email:email` we can use just `:email` for simplicity.


>Thanks! I'll change it soon.

Screenshot 2019-11-13 at 15 23 54

PR #88, opened 2019-11-13
feat(nav): add see more examples button

williammartin:


williammartin:
I don't love these tests, but they are what they are for now. They could use to be made explicitly relating to the navbar in the studio page itself.

Arinono:
## Acceptance
**When** I am navigate to an example
**Then** I see a button that says `See More Examples`

Screenshot 2019-11-13 at 15 13 52

PR #91, opened 2019-11-13
bug(tips): fix locations for counter example

williammartin:


williammartin:
### Acceptance

**When** I navigate to the counter example
**Then** I see tips in the right location

image

LGTM

PR #95, opened 2019-11-13
refactor(function): update mail template + sender

Arinono:


Arinono:
Pushed on dev by mistake after acceptance 😓 .
Closing

PR #101, opened 2019-11-13
feat(welcome): match descriptions to landing page

williammartin:


williammartin:
### Acceptance

**When** I look at the example cards
**Then** I see the text descriptions are the same as those at `storyscript.io` landing page

image

PR #109, opened 2019-11-14
[PoC] feat(editor): add language client connection logic

wilzbach:
TODO:
- [x] deploy SLS inside the Storyscript cluster (available as Docker image here: https://hub.docker.com/repository/docker/storyscript/sls)
- [x] improve SLS (see e.g. https://github.com/storyscript/sls/projects/3)


src/components/MonacoEditor.vue
```diff
@@ -66,7 +69,11 @@ export default class MonacoEditor extends Vue {
       ...this.options,
       value: this.value,
       theme: this.theme,
-      language: this.language
+      language: this.language,
+      glyphMargin: true,
+      lightbulb: {
+        enabled: true
+      }
```
wilzbach: These don't have an effect yet as SLS doesn't emit the respective events yet, so I guess we can remove them for now and re-enable only once needed.

stevepeak:
Screen Shot 2019-12-11 at 07 01 12

Can we load auto-complete without <Control>+<Space>?

----

Screen Shot 2019-12-11 at 07 02 14

Can the default state be different? Perhaps we want to
1. Show popular services
2. Predict what you may want to use next

----

Screen Shot 2019-12-11 at 07 06 35

This should be the default state, not showing the full details in another panel. When you click the (i) icon we can open the service in the Library side panel.

Screen Shot 2019-12-11 at 07 07 46

Can we suggest previous variable when we know it should be a `<string>`?

Screen Shot 2019-12-11 at 07 08 43

1. Remove the `E0129: `
2. Where are these errors listed? I would like to write more docs around them to when users click on our error cards they can get more information about what the error is concerning

Screen Shot 2019-12-11 at 07 11 00

What are `abc` suggestions?

src/components/MonacoEditor.vue
stevepeak: Can you link me to examples/docs about what these features provide to the user?

wilzbach:
> Can we load auto-complete without +?

That's entirely up to the editor, so sure. I added it to the TODO.

> Can the default state be different? Perhaps we want to
> Predict what you may want to use next

Yes, e.g. one simple improvement: https://github.com/storyscript/sls/issues/162

> When on an empty line: maybe suggest more generic things like "Create a variable, call a function or call a service". Then it could auto-complete these concepts. For example, call a service would create <result> = <service> <action>

Interesting idea. I'm not sure how this would work out in practice, but I created an issue at SLS: https://github.com/storyscript/sls/issues/172

> This should be the default state, not showing the full details in another panel. When you click the (i) icon we can open the service in the Library side panel.

Definitely, but I'm not too concerned about the visual representation of the SLS output at the moment as the Studio is basically free to do what it wants as long as it receives a "correct" JSON response payload. 

> Can we suggest previous variable when we know it should be a <string>?

Actually this should/will work (see e.g. https://github.com/storyscript/sls/pull/171 as a test). 
I created a new issue for filtering down based on the type: https://github.com/storyscript/sls/issues/170

> Remove the E0129:

The idea of the error code was that it is easy to google the error (or at least will be).
Anyhow, the Storyscript compiler already produces the codes to us for free, but we can easily strip them away (added it to the TODO of this PR).

> Where are these errors listed? I would like to write more docs around them to when users click on our error cards they can get more information about what the error is concerning

https://github.com/storyscript/storyscript/blob/master/storyscript/ErrorCodes.py

See also e.g. https://github.com/storyscript/storyscript/issues/653

> What are abc suggestions?

This happens when SLS crashes and VSCode does its own word selection based on all the words it finds in the document. We can definitely disable this.
While of course the underlying issue should/will be fixed, SLS already has a "safe" fallback on a crash during completion:

https://github.com/storyscript/sls/blob/2ed31da2406a3a1a8ee6f7a0d930bafb2be14d2a/sls/completion/complete.py#L48-L50

Though we might want to add a similar logic to the client as well as a crash of the server (or the connection) might always happen. I added it to the TODO of this PR.

> Auto-completing events

Event completion isn't supported yet, but I was/am already working on a PR for it.
The issue is https://github.com/storyscript/sls/issues/166.

> Let's only suggest services that have actions w/ events

Yes, on the roadmap for 0.5.0 as well (https://github.com/storyscript/sls/issues/166).
Similarly, in assignments SLS will only show actions and not events soon (https://github.com/storyscript/sls/issues/164).

JeanBarriere:
- [x] implemented with new architecture
- [x] fix tests
- [ ] add unit tests for sls plugin

src/editor/core/types/editor.ts
```diff
@@ -0,0 +1,68 @@
...
+  folding: false,
+  autoIndent: true,
+  cursorBlinking: 'blink',
+  cursorWidth: 1,
+  lineNumbers: 'off',
+  glyphMargin: true,
+  lightbulb: {
+    enabled: true
+  }
+}
```
JeanBarriere: new options for SLS

src/editor/core/types/editor.d.ts
```diff
@@ -0,0 +1,38 @@
...
+export class EditorLang {
+  name: string
+  extension: string
+  icon: string
+}
+
+export interface EditorPlugin {
+  attach: (editor: IStandaloneCodeEditor) => void
+  detach: () => void
+}
```
JeanBarriere: EditorPlugin interface

src/editor/plugin-sls/actions/enterIndent.ts
```diff
@@ -0,0 +1,47 @@
...
+        const value = lines.join('\n')
+        ed.getModel()!.setValue(value)
+        const newPos = new monaco.Position(lineIndex + 2, indent.length + 1)
+        ed.setPosition(newPos)
+
+        return Promise.resolve()
+      })
+    }
+  }
+}
```
JeanBarriere: enterIndent action

src/editor/plugin-sls/actions/index.ts
```diff
@@ -0,0 +1 @@
+export { enterIndentAction } from './enterIndent'
```
JeanBarriere: exporting the list of actions

src/editor/plugin-sls/index.ts
```diff
@@ -0,0 +1,98 @@
...
+  createConnection,
+  ErrorAction,
+  MonacoLanguageClient,
+  MonacoServices
+} from 'monaco-languageclient'
+import ReconnectingWebSocket from 'reconnecting-websocket'
+import 'setimmediate' // required by vscode-jsonrpc
+import { listen, MessageConnection } from 'vscode-ws-jsonrpc'
+import { EditorPlugin, IStandaloneCodeEditor } from '&/editor.d'
+// import { enterIndentAction } from '@editor-plugin-sls/actions'
```
JeanBarriere: import actions

src/editor/plugin-sls/index.ts
```diff
@@ -0,0 +1,98 @@
...
+
+const normalizeURL = require('normalize-url')
+
+export class SLS implements EditorPlugin {
```
JeanBarriere: SLS implements EditorPlugin

src/editor/plugin-sls/index.ts
```diff
@@ -0,0 +1,98 @@
...
+  private languageClient: MonacoLanguageClient | undefined
+
+  attach (editor: IStandaloneCodeEditor) {
+    this.connect(editor)
+    // editor.addAction(enterIndentAction(this))
+  }
+
+  detach () {
+    this.disconnect()
+  }
```
JeanBarriere: attach and detach for connection/disconnection and adding actions

src/editor/plugin-sls/types/sls.d.ts
```diff
@@ -0,0 +1,10 @@
+import { MonacoLanguageClient } from 'monaco-languageclient'
+import { EditorPlugin, IStandaloneCodeEditor } from '&/editor.d'
+
+export declare class SLS implements EditorPlugin {
```
JeanBarriere: declare class so we can import in actions

src/app/views/Editor/EditorPane.vue
```diff
@@ -0,0 +1,78 @@
...
+  </div>
+</template>
+
+<script lang="ts">
+import { Component, Vue, Watch } from 'vue-property-decorator'
+import { Getter } from 'vuex-class'
+import MonacoEditor from '@editor/MonacoEditor.vue'
+import { Editor, Languages } from '&/editor'
+import { IStorySample } from '&/StorySample'
+import SLS from '@editor-plugin-sls'
```
JeanBarriere: import SLS plugin

src/app/views/Editor/EditorPane.vue
```diff
@@ -0,0 +1,78 @@
...
+export default class EditorPane extends Vue {
+  private editors: Array<Editor> = []
+
+  @Getter('getPayload')
+  private payload!: IStorySample
+
+  @Watch('payload', { immediate: true })
+  private onPayloadUpdate () {
+    this.editors.splice(0, this.editors.length)
+    this.editors.push(new Editor(Languages.Storyscript, this.payload.code, { readOnly: false }, [new SLS()]))
```
JeanBarriere: SLS is added to the Editor plugins list

src/editor/plugin-sls/index.ts
```diff
@@ -74,7 +80,7 @@ export class SLS {
     if (url.protocol === 'https:') {
       url.protocol = 'wss:'
     } else {
-      url.protocol = 'ws:'
+      url.protocol = 'wss:'
```
wilzbach: This doesn't look correct.

PR #119, opened 2019-11-18
fix(navbar): fix navbar routing issue

Arinono:
## Acceptance

**Given** I am on an example and click the SS logo to take me to the welcome page
**When** I press the back button
**Then** I should see the nav bar as expected for being in the studio (with publish etc)

### DIFF
--- a/package.json
+++ b/package.json
@@ -98,7 +98,7 @@
     }
   },
   "lint-staged": {
-    "*.{js,vue}": [
+    "*.{js,vue,ts}": [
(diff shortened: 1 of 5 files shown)


Arinono:
## Acceptance

**Given** I am on an example and click the SS logo to take me to the welcome page
**When** I press the back button
**Then** I should see the nav bar as expected for being in the studio (with publish etc)

Screen Recording 2019-11-18 at 16 19 18

PR #124, opened 2019-11-19
feat: add deepspeech icon and update the zoom example

Arinono:
## Acceptance

**When** I look at the zoom example
**Then** I see a DeepSpeech icon instead of Machine Box

_Notes: maybe use Mozilla as they created deepspeech_

--- 

Screenshot 2019-11-19 at 14 20 31


JeanBarriere:
**When** I look at the zoom example from the welcome page
image

**Then** I see a DeepSpeech icon instead of Machine Box
image

PR #130, opened 2019-11-21
feat(archi): remove k8s and story tiles + col displaying

Arinono:
## Acceptance

**Given** I navigate to an example
**Then** I no longer see the kubernetes or storyscript sections of the architecture diagram

**And When** I look at an example that has many services
**Then** I see them tiling from top down (i.e if there are 3 then 1 is on the bottom row) rather than expanding horizontally when there are more than 2

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,1 @@
+* @storyscript/frontend 
--- a/src/views/Studio/Architecture.vue
+++ b/src/views/Studio/Architecture.vue
@@ -12,107 +12,63 @@
       Architecture
     </s-text>
     <perfect-scrollbar class="bg-gray-10 p-8 max-h-xs">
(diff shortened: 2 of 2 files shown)


stevepeak:
lgtm 👍 

Screen Shot 2019-11-21 at 11 55 34

williammartin:
My acceptance on this story was grammatically very bad. What I intended to convey was that there should be 2 boxes per row, so that we don't get a scroll bar after 3 services.

Arinono:
@stevepeak @williammartin Do you prefer the actual version (1 per row), or this ?
Screenshot 2019-11-21 at 12 31 29

williammartin:
I'll defer to @stevepeak on this as he is starting to pick up PMing duties here.

Arinono:
Ok, I'll merge what we have right since he accepted, and there already un update on that part

PR #133, opened 2019-11-21
style(archi): new grid layout

Arinono:
## Acceptance

**Given** I navigate to an example
**Then** the Services stack in a grid instead of as a list
**And** each service is equal in size 200px wide

---

Not exactly 200px due to responsive issue (iPad), but still smaller cards


stevepeak:
Screen Shot 2019-11-21 at 17 15 27
lgtm 👍

stevepeak:
Screen Shot 2019-11-21 at 17 15 27
lgtm 👍

PR #134, opened 2019-11-21
feat(share): add share btn + copy to clipboard

Arinono:
- [x] add share button
- [x] copy to clipboard
- [x] unit test
- [x] e2e test


stevepeak:
@Arinono Let's not close the model window when you click copy. I felt it was an unexpected behavior.

I also informed Ayda about the design changes to this window, but we will save that for another pull.

Arinono:
@stevepeak I made the requested change.

PR #151, opened 2019-11-26
Feat/persistent tips show

JeanBarriere:
- [x] add vuex-persistedstate dependency
- [x] add Tips vuex module and store shownSamples ids
- [x] change Studio index to take isIntro boolean from store and update it when it's done
- [x] add unit tests

___
### Acceptance
**When** I end the tutorials
**Then** they do not automatically appear again


src/store/index.ts
```diff
@@ -1,10 +1,16 @@
...
+import { version } from '@/../package.json'
 
 Vue.use(Vuex)
 
 export default new Vuex.Store({
   modules,
-  strict: process.env.NODE_ENV !== 'production'
+  strict: process.env.NODE_ENV !== 'production',
+  plugins: [createPersistedState({
+    key: `studio-v${version.split('.')[0]}`, // keeping storage away from potential breaking changes
```
stevepeak: But what about tips? Can you skip this forever? I read this as it will show tips if we upgrade studio versions

JeanBarriere: It will show them in case of a major version update (e.g. there is a breaking change in the studio)

PR #153, opened 2019-11-27
[WIP] Add toolbar

adnrs96:
- [x] Add support for drop component to open up on left/right
- [x] Add toolbar as specified by design.
- [x] Add tests
- [ ] Fix/add unit tests for drop.vue


src/assets/styles/custom/_drop.scss
```diff
@@ -59,6 +59,42 @@
...
+    &.left-full.bottom-0 {
+      @apply ml-4 -mb-2;
+
+      &:before {
+        bottom: .75rem;
+        left: -.2rem;
+      }
+    }
+
+    &.right-full.top-0 {
```
adnrs96: I realise the naming here isn't ideal. I actually would like to change naming sceme altoghter here to something that doesn't depend upon the values inside, given this is being used by drop to apply css and these classes are being added using logic that requires them to be coherent in naming

adnrs96:
Closing this in favour of #157

PR #157, opened 2019-11-29
Add tool bar as specified by design

adnrs96:
- Add capability to drop component to be rendered on left/right
- Improved logic for automatically placing drop component 
- Add Toolbar according to design 
- Add unit tests
- Add integration tests


stevepeak:
Do not merge. In light of new requirements we may change the toolbar orientation.

PR #169, opened 2019-12-05
fix(routing): fix issue when redirecting to /welcome

Arinono:
- [x] Add missing unit test coverage

## Acceptance

**When** I navigate to https://studio.storyscript.io/
**And** I click Explore on any of the examples
**And** I click the back button
**And** I click the back button again
**Then** I should be shown the example I was previously at correctly

---

>**Then** I should be shown the example I was previously at correctly

I was not able to do that, instead I rechecked the route, so the user at least stays on `/welcome`

### DIFF
--- a/src/views/index.vue
+++ b/src/views/index.vue
@@ -9,13 +9,22 @@
 </template>
 
 <script lang="ts">
-import { Component, Vue } from 'vue-property-decorator'
+import { Component, Vue, Watch } from 'vue-property-decorator'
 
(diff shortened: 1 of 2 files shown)


williammartin:
### Acceptance

**When** I navigate to the welcome page

image

**And** I click `Explore` on any of the examples

image

**And** I click the back button

image

**And** I click the back button again
**Then** I should be navigated to the page I was at before

image

PR #178, opened 2019-12-09
fix(navbar): fix new from template button style

Arinono:
Closes #172


src/views/RegisterInterest.vue
```diff
@@ -110,24 +110,17 @@
...
-          color="text-gray-100"
-          class="px-3 py-2 whitespace-no-wrap"
-          clickable
-          @click="$refs.registerInterestModal.show()"
-        >
-          New from template
-        </s-text>
-      </div>
+      <s-gradient-border-button
+        secondary
```
JeanBarriere: why `secondary` ?

Arinono: Because on light theme, the default color is black. So the secondary is the opposite (white background)

PR #182, opened 2019-12-10
Feat/share feedback

adnrs96:
### Acceptance

**When** I navigate to any Story
**Then** I see the 😃share feedback button in the toolbar

**When** I click the share feedback button
**Then** a popup appears asking the user for their feedback
**And** The only requirement is that one field is filled in (no blank entries allowed to submit)

**When** I click submit
**Then** Feedback is collected via Netlify built-in form
**And** the popup shows a Thank You and closes after 2500ms

Work:

- Add selectPills component
- make input component capable of using textarea tag
- add feedback modal
- add unit tests
- add e2e tests

Pending: 

Modal from upstream component has weird padding and I can't seem to fix the issue with content compressed into little space in a huge modal

**closes #183**


adnrs96:
e2e test failure occurs on dev as well for me.
Something is up with register-interest.e2e.ts but I am not sure :confused:

PR #188, opened 2019-12-11
Fix selectpills behaviour of retaining selection after modal closes.

adnrs96:
- Use v-model construct instead of custom event
- Adapt unit tests to new code

### DIFF
--- a/tests/unit/components/Inputs/SelectPills.spec.ts
+++ b/tests/unit/components/Inputs/SelectPills.spec.ts
@@ -33,9 +33,9 @@ describe('SelectPills.vue', () => {
         }
       })
 
-      const vm = pills.vm as any
-      vm.onPillSelected()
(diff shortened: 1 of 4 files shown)


tests/unit/components/Inputs/SelectPills.spec.ts
```diff
@@ -33,9 +33,9 @@ describe('SelectPills.vue', () => {
         }
       })
 
-      const vm = pills.vm as any
-      vm.onPillSelected()
-      expect(pills.emitted().pillSelectionChange).toBeTruthy()
+      const option = pills.findAll('label').at(0)
+      option.trigger('click')
```
adnrs96: Not sure if there's a better way to do this since now vm does not have a function that triggers events?

Arinono: It the test passes, it doesn't really matter. It was a "fix" for a dependency update

PR #192, opened 2019-12-11
feat(library): add library pane

Arinono:
## Todo
- [x] remove toolbar
- [x] move Storyscript Logo in Navbar
- [x] add library pane
- [x] add toggle lib pane behavior
- [x] unit tests
- [x] ~E2E tests~ I'll wait that we merge other tasks (new Editor Pane + remove fake components)

Blocked by #190 
Closes #191 

---

### Acceptance

**When** I navigate to a story
**Then** I no longer see the toolbar on the left side
**And** I see a pane on the left side of the code titled `Library`
**And** It has the content from the design

**And** The storyscript logo has moved to the blue header

---

**Note** The pane does not have to be collapsable (as is in the design right now)

PR #193, opened 2019-12-11
feat(editor): add editor pane views + editor classes + tests

JeanBarriere:
- [x] add new editor pane
- [x] new design
- [x] new editor types
- [x] unit tests
- [ ] add temporary route for testing
___
- [x] **blocked by #192 for proper routing**
___
### Acceptance

**When** I navigate to a Story
**Then** I see the gray background and code editor as a pane in the middle.

**Default state** code line numbers are not visible
**When** I hover or interact with a block of code
**Then** The line numbers fade-in

-----
The right gutter of the code editor is removed entirely.
The color of the line numbers is changed according to the design.


stevepeak:
Screen Shot 2019-12-11 at 18 41 12


@JeanBarriere  the design of the code editor is not that of the design. Am I missing something here? Cache issue?

JeanBarriere:
Not cache, you got the preview while I was adding my changes to the current view

Arinono:
> **Then** I see the gray background and code editor as a pane in the middle.

Because it's unsure with the first screenshot, the line number are hidden when not focused:
Screenshot 2019-12-11 at 19 03 20

PR #196, opened 2019-12-12
Feat/gh login popover

adnrs96:
This is the first PR in a possible series of PR to add gh-login :)

- Add GH-signup-popover from design
- Add Invite-signup-popover from design
- Add unit tests.


adnrs96:
Mismatch with design currently includes

- Different github icon
- Missing divider
- Possibly too much padding on modal

Padding also prevents me from adding the divider or maybe their is a way to add it even though their is padding on parent modal component?

src/app/views/Login.vue
```diff
@@ -0,0 +1,137 @@
...
+              weight="semibold"
+              class="text-white"
+            >
+              Sign up with Github
+            </s-text>
+            <s-icon
+              icon="github-f"
+              color="text-white"
+              class="flex flex-1 justify-end"
+              height="20"
```
Arinono: Design says 24x24

src/app/views/Login.vue
```diff
@@ -0,0 +1,137 @@
...
+            />
+          </div>
+        </s-button>
+
+        <div
+          id="beta-form"
+          class="mt-6 w-full"
+          name="beta"
+        >
+          <s-labeled-input
```
Arinono: Maybe it's part of a previous design, but the render has two `mb-2` classes here. One on the label, one on the `s-text`. So the spacing between `Email` and the input is to big.

src/app/views/Login.vue
```diff
@@ -0,0 +1,137 @@
...
+            shadow
+          >
+            <div
+              slot="label"
+              class="font-medium text-gray-80"
+            >
+              Email
+            </div>
+          </s-labeled-input>
+          <s-text
```
Arinono: You can use `s-message` here, this way you don't have to take cake of the font size/weight, ...

[See](https://components.storyscript.io/#/Components?id=message)

src/app/views/Invite.vue
```diff
@@ -0,0 +1,115 @@
...
+            shadow
+          >
+            <div
+              slot="label"
+              class="font-medium text-gray-80"
+            >
+              Invite a friend
+            </div>
+          </s-labeled-input>
+          <s-text
```
Arinono: You can use `s-message` here, this way you don't have to take cake of the font size/weight, ...

[See](https://components.storyscript.io/#/Components?id=message)

src/app/views/Login.vue
Arinono: Screenshot 2019-12-12 at 14 31 04
Screenshot 2019-12-12 at 14 31 24

src/app/views/Login.vue
adnrs96: hmm, my bad. I kinda manually selected the icon and saw this
image

As a result I used 20x20

Now you mentioned that its 24x24 and I went looking and realised that I didn't really knew how to use figma to find stuff like this. Now I figured that I need to use the left panel to select the icon named github-f and then view props.

I will update this. Thanks for the review.

src/app/views/Login.vue
adnrs96: Well this comes from LabeledInput. I added a commit removing this.

src/app/views/Login.vue
adnrs96: addressed this via editing commit

src/app/views/Invite.vue
adnrs96: addressed this via editing commit

src/app/views/Login.vue
adnrs96: addressed this via editing commit

PR #199, opened 2019-12-12
feat(editor): add language client connection logic

wilzbach:
> As I fucked a bit the history tree when I rebased, [...]

CC @JeanBarriere - this is rebased properly to master and as my intermediate explorative commits aren't meaningful, I squashed them to one.

See also: https://github.com/storyscript/studio/pull/109

TODO:
- [x] add unit test for SLS plugin
- [x] remove indent action
- [ ] remove error code from diagnostics (https://github.com/storyscript/studio/pull/109#issuecomment-564396063)
- [x] disable fallback suggestion on SLS crash (https://github.com/storyscript/studio/pull/109#issuecomment-564396063)


src/editor/plugin-sls/actions/enterIndent.ts
```diff
@@ -0,0 +1,47 @@
+import * as monaco from 'monaco-editor/esm/vs/editor/editor.api'
+import { SLS } from '&/sls/types/sls.d'
+
+/**
+Overloads the 'enter' key in the editor for a custom indentation behavior.
+*/
+// TODO: transform into experimental LSP Feature and add as capability
+export function enterIndentAction (completion: SLS) : monaco.editor.IActionDescriptor {
```
wilzbach: This is only a PoC (+ the feature isn't deployed at SLS yet), so I would suggest removing this fully and adding it in a separate PR, s.t. we can take a look at it separately.

src/editor/plugin-sls/index.ts
```diff
@@ -0,0 +1,98 @@
...
+        }
+      }
+    })
+  }
+
+  private createWebSocket (url: URL): WebSocket {
+    if (url.protocol === 'https:') {
+      url.protocol = 'wss:'
+    } else {
+      url.protocol = 'wss:'
```
wilzbach:
```suggestion
      url.protocol = 'ws:'
```

src/editor/plugin-sls/types/sls.d.ts
```diff
@@ -0,0 +1,10 @@
+import { MonacoLanguageClient } from 'monaco-languageclient'
+import { EditorPlugin, IStandaloneCodeEditor } from '&/editor.d'
+
+export declare class SLS implements EditorPlugin {
+  attach: (editor: IStandaloneCodeEditor) => void
+  detach: () => void
+  connect: (editor: IStandaloneCodeEditor) => void
+  getLanguageClient: () => MonacoLanguageClient
+  disconnect: () => void
+}
```
wilzbach: I don't see the purpose of an interface if you don't use it.

tests/unit/components/MonacoEditor.spec.ts
```diff
@@ -47,6 +48,19 @@ describe('MonacoEditor.vue', () => {
...
+    vm.initMonaco()
+    expect(view.vm).toHaveProperty('plugins', [plugin])
+  })
+
+  it('should not have plugins', () => {
+    view.setProps({ plugins: undefined })
+    vm.initMonaco()
+    expect(view.vm).toHaveProperty('plugins', undefined)
+  })
+
```
wilzbach: Should we setup a few bare test for completion support?

If network access is an issue, we could easily mock the websocket server for testing.

Background infos
-----------------------

I know that a lot of communication currently happen via the WS:

image

(that's because SLS is already subscribed to hover events -I already created https://github.com/storyscript/sls/issues/173 for this).

But a mocked server would only need to do the following:
1) receive client capabilities payload (can just ignore it) and returns server capabilities payload

```json
{"jsonrpc": "2.0", "id": 0, "result": {"capabilities": {"completionProvider": {"resolveProvider": false, "triggerCharacters": [".", "(", " "]}, "hoverProvider": true, "documentFormattingProvider": true, "textDocumentSync": {"openClose": true, "change": 1}}}}
```

We can even remove the hoverProvider and documentFormattingProvider for a simple test.

2) client sends didChange to server (no mock here - we want to test this)

```json
{"jsonrpc":"2.0","id":4,"method":"textDocument/completion","params":{"textDocument":{"uri":"inmemory://model/1"},"position":{"line":3,"character":3},"context":{"triggerKind":1}}}
```

and server replies with 

```json
{"jsonrpc": "2.0", "id": 4, "result": {"isIncomplete": false, "items": [{"label":"rabbitmq","kind":2,"detail":"An OMG compliant container to communicate with RabbitMQ, a message broker","insertTextFormat":1,"textEdit":{"range":{"start":{"line":3,"character":2},"end":{"line":3,"character":3}},"newText":"rabbitmq "},"sortText":"40-rabbitmq", "documentation": {"value": "..."}}]}}
```

src/editor/plugin-sls/types/sls.d.ts
JeanBarriere: it's just the declaration, imported in the action so there is no circular import

JeanBarriere:
I'm not sure we should mock the sls server

wilzbach:
> I'm not sure we should mock the sls server

Well, imho we should definitely test whether completions are actually rendered or not (i.e. an integration test). I think if you use the live server this might lead to unrelated failures (i.e. when the completion improves), but that's just my opinion.

JeanBarriere:
> we should definitely test whether completions are actually rendered or not (i.e. an integration test)
Well, we can make an e2e with the monaco editor for this, but the value rendered should be tested on the sls directly as it's not really studio-side to care about what value is returned for which input, and this can break easily

> I think if you use the live server this might lead to unrelated failures
True, that's why we should definitely embed it into a local cluster to do these tests I guess

package.json
```diff
@@ -20,6 +20,12 @@
   "dependencies": {
     "@storyscript/vue": "3.3.0",
     "core-js": "3.5.0",
+    "jest-canvas-mock": "2.2.0",
```
Arinono: Is it possible to pin these dependencies before Renovate does ?

JeanBarriere: done

yarn.lock
```diff
@@ -13774,6 +13831,49 @@ vm-browserify@^1.0.1:
...
 
+vscode-jsonrpc@^4.1.0-next:
+  version "4.1.0-next.3"
+  resolved "https://registry.yarnpkg.com/vscode-jsonrpc/-/vscode-jsonrpc-4.1.0-next.3.tgz#05fe742959a2726020d4d0bfbc3d3c97873c7fde"
+  integrity sha512-Z6oxBiMks2+UADV1QHXVooSakjyhI+eHTnXzDyVvVMmegvSfkXk2w6mPEdSkaNHFBdtWW7n20H1yw2nA3A17mg==
+
+vscode-jsonrpc@^5.0.0-next.3:
+  version "5.0.0-next.4"
+  resolved "https://registry.yarnpkg.com/vscode-jsonrpc/-/vscode-jsonrpc-5.0.0-next.4.tgz#bbe1f7b4876eefe3e93bffd93672125506698d59"
+  integrity sha512-Tos3tXP62ZTB9WowWwhvfVNdu1mEwQF/j7DqJuVL4QKhk311gH+mda0PZpG95LWyh5CCRpHMns4vNmMgZQrvXQ==
```
wilzbach: I guess ideally we should pull in one version only.
Though it looks like vscode-ws-jsonrpc (https://github.com/TypeFox/vscode-ws-jsonrpc/blob/master/package.json) requires v4.

PR #200, opened 2019-12-13
chore(InfosPane): make infospane fixed.

adnrs96:
Make the suggestions/errors text fixed in the infos pane.

I am not sure about how this should be done correctly but this way did work for my screen. Not sure if this is the correct thing to do which would work on all screens.
context: `inherit` uses px values and not computed values??

story: https://www.pivotaltracker.com/story/show/170277014

### DIFF
--- a/src/app/views/Editor/InfosPane.vue
+++ b/src/app/views/Editor/InfosPane.vue
@@ -4,7 +4,7 @@
       p="4"
       weight="medium"
       align="center"
-      class="text-gray-50 mt-3"
+      class="text-gray-50 mt-3 sticky top-3"
     >
       0 Suggestions, 0 Error
     </s-text>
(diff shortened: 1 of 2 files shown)


adnrs96:
Thanks @Arinono for feedback on chat. Updated this.

src/app/views/Editor/InfosPane.vue
```diff
@@ -1,13 +1,15 @@
...
   <div id="infos-pane">
-    <s-text
-      p="4"
-      weight="medium"
-      align="center"
-      class="text-gray-50 mt-3"
-    >
-      0 Suggestions, 0 Error
-    </s-text>
+    <div class="sticky top-3">
```
Arinono: @adnrs96 You can remove this div and put the classes directly on the `s-text`, right ?x

adnrs96: sure

PR #201, opened 2019-12-13
feat(tabs): add tabs + base views

Arinono:
Closes #197


tailwind.config.js
```diff
@@ -151,12 +152,13 @@ module.exports = {
...
         min: '.125rem',
         1: '.25rem',
+        3: '.75rem',
         half: '50%',
         full: '100%'
       },
       maxWidth: {
         xxs: '10rem',
-        pane: '16rem'
+        pane: '19rem'
```
JeanBarriere: :thinking:

tailwind.config.js
```diff
@@ -151,12 +152,13 @@ module.exports = {
       inset: {
         min: '.125rem',
         1: '.25rem',
+        3: '.75rem',
```
JeanBarriere: :thinking:

tailwind.config.js
```diff
@@ -139,7 +139,8 @@ module.exports = {
         'max-content': 'max-content',
         '2full': '200%',
         14: '3.5rem', // 56px
-        22: '5.5rem' // 88px
+        22: '5.5rem', // 88px
+        76: '19rem' // 304px
```
JeanBarriere: :thinking:

src/app/router.ts
```diff
@@ -18,13 +18,30 @@ export default new Router({
...
+              path: 'events',
+              name: 'story-events',
+              component: () => import('@app/Events/index.vue')
+            },
+            {
+              path: 'services',
+              name: 'story-services',
+              component: () => import('@app/Services/index.vue')
+            }
+          ]
```
JeanBarriere: to prevent potential errors:
- add the `/` in the alias as well
- remove the name `studio` from the parent route
- add a child with no path and the name `studio`

tailwind.config.js
Arinono: This was made during a call with Steve and Ayda, when we discussed about the editor size. When doing the live changes I needed changed it to try, and we validated this size (300px)

tailwind.config.js
Arinono: This is part of the fix about the "0 Suggestions, 0 Error" text that was scrollable. When Aditya added the `fixed top-3`.

See https://github.com/storyscript/studio/pull/200/files

tailwind.config.js
Arinono: Again this is the new size for the Library pane. 300px is bigger than the maximum size offered by tailwind (16rem, 256px). And the library pane is a fixed with, we will probably never have to resize it (but instead hide it)

src/app/router.ts
Arinono: Sure, will do, thanks 😊

PR #204, opened 2019-12-16
feat(login): add login components and graphql support

JeanBarriere:
- [x] add login modal (#196) 
- [x] add graphql dependencies (#205)
- [x] add avatar in navbar (#210)
- [x] add login process (#210)
- [x] e2e tests on login process (#270)


src/app/views/Invite.vue
```diff
@@ -0,0 +1,114 @@
...
+  }
+})
+export default class Invite extends Vue {
+  private email = ''
+  private emailError = ''
+
+  @Watch('email')
+  private onEmailHandler () {
+    if (
+      /^(([^<>()\\[\]\\.,;:\s@\\"]+(\.[^<>()\\[\]\\.,;:\s@\\"]+)*)|(\\".+\\"))@(([^<>()[\]\\.,;:\s@\\"]+\.)+[^<>()[\]\\.,;:\s@\\"]{2,})$/.test(
```
wilzbach: Big no on this. No one will understand this in two weeks. That's the perfect reason to use a package like e.g. https://www.npmjs.com/package/email-validator

src/app/views/Login.vue
```diff
@@ -0,0 +1,136 @@
...
+  }
+})
+export default class Signup extends Vue {
+  private email = ''
+  private emailError = ''
+
+  @Watch('email')
+  private onEmailHandler () {
+    if (
+      /^(([^<>()\\[\]\\.,;:\s@\\"]+(\.[^<>()\\[\]\\.,;:\s@\\"]+)*)|(\\".+\\"))@(([^<>()[\]\\.,;:\s@\\"]+\.)+[^<>()[\]\\.,;:\s@\\"]{2,})$/.test(
```
wilzbach: And there's duplication (!)

src/internal/plugins/vue-apollo.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  const apolloProvider = new VueApollo({
+    defaultClient: apolloClient,
+    defaultOptions: {
+      $query: {
+        fetchPolicy: 'cache-and-network'
+      }
+    },
+    errorHandler (error: any) {
+      console.log('%cError', 'background: red; color: white; padding: 2px 4px; border-radius: 3px; font-weight: bold;', error.message)
```
wilzbach: Why not simply `console.error`?

Arinono: I don't have the exacte answer, but I guess we just want the error message and not the whole stack trace 🤷‍♂

src/app/views/Invite.vue
Arinono: I disagree a little bit with you on that. I can't read a RegExp like that, but in a method called `onEmailHandler` I'll always know what's this RegExp is doing

src/internal/plugins/vue-apollo.ts
wilzbach: `console.error(error.message)`

src/app/views/Invite.vue
wilzbach: >  I can't read a RegExp like that

Yup and there are almost no tests.
Also, this RegExp is very much incorrect.
Two quick examples:

image

image

src/internal/plugins/vue-apollo.ts
Arinono: Yes I know, what I was implying is that `console.error` is also giving the full stack trace, and `console.log` is not. I don't know if this is the reason, but this is the only difference I see.

wilzbach: > console.error is also giving the full stack trace

That's optional and up to the IDE/browser on how to display. For example, by default, Chrome doesn't show the full stack trace, but let's you toggle it (another reason for `console.error`).

There's also the point that e.g. using custom CSS in log messages doesn't work well on the CLI nor with third-party logging tools like sentry.

src/app/views/Invite.vue
JeanBarriere: :+1: we should have some utils to merge these and prevent code duplication, but adding a dependency for a simple regex check is a bit overkill to me.

src/internal/plugins/vue-apollo.ts
JeanBarriere: This is generated by the vue cli plugin apollo and I honestly didn't checked:
https://github.com/Akryum/vue-cli-plugin-apollo/blob/master/generator/templates/vue-apollo/default/src/vue-apollo.js#L73

JeanBarriere: The answer I guess is to provide a log and not an Error in the console, as graphql errors are to be interprated differently as a console.error message

JeanBarriere: Screenshot from 2019-12-19 12-24-44
First one is a browser error, as Second one is just a log, removable, or to be dispatched on another layer (sentry or whatsoever)

src/app/views/Invite.vue
wilzbach: It's not simple as can be seen above. The first example that I tried was a false positive.

JeanBarriere: I'm not saying we should keep the regex as it is, but in any case the `email-validator` package is just implementing another regex

PR #212, opened 2019-12-19
Feat/sls error

adnrs96:
**Acceptance**
Given I navigate to a Story I can edit.
When the SLS presents an error event
Then an "Error Card" appears showing the issue to the side of the code
And the "Error card" is anchored to the line during scroll events

When my text cursor is in the bounds of the culprit code
Then the "Error Card" moves slightly left with a gentle animation

When my text cursor leaves the bounds of the culprit code
Then the "Error Card" returns to its original position with a gentle animation

This removes the behavior of "Peek Problem" (see comment below)

Work:
- [x] Add interface, state, getters and mutations for InfosPane vuex store to get error data.
- [x] Add Plugin for reporting SLS errors (this works by swaping out a function from monaco)
- [x] Add error cards and connect error plugin with infospane to show error cards.
- [x] Remove default error reporting and peek behaviour
- [ ] Make errors disappear/reappear with cool animation. (time when error cards first appear)
- [ ] Make errorcards fade out and fade back in on scrollevents when cards go out of scope.
- [x] Add/Fix unit tests
- [x] Add/fix e2e tests


src/app/views/Editor/InfosPane.vue
```diff
@@ -6,17 +6,45 @@
...
+    <div
+      v-for="error in errors"
+      :key="`error-brick-${error.line}-${error.msg}`"
+      :style="{
+        marginTop: error.relativeTop + 26 + 'px',
+        marginLeft: cursorLine === error.line? '-2rem': 'unset'
+      }"
+      class="transition-all-fastest"
+    >
+      <div class="flex flex-row absolute w-fit-content min-w-brick max-w-pane bg-white border border-solid border-red-30 rounded-lg shadow-sm p-1">
```
JeanBarriere: please use https://tailwindcss.com/docs/max-width instead instead of adding a min-w-brick, for better responsive behavior

src/app/views/Editor/InfosPane.vue
```diff
@@ -6,17 +6,45 @@
...
+        <div
+          class="rounded-2xl border-red-60 border-2 mr-2 h-8"
```
JeanBarriere: also no need to add a 2xl border radius, the width is not big enough to see a difference with a normal border radius

src/editor/plugin-error/index.ts
```diff
@@ -0,0 +1,37 @@
...
+interface ISetModelMarkers {
+    (model: editor.ITextModel, owner: string, markers: editor.IMarkerData[]): void
+}
+
+export class StoryscriptError implements EditorPlugin {
+    private readonly monacoEditor: typeof editor
+    private readonly setModelMarkersOriginal: ISetModelMarkers
+
+    constructor (monacoEditor: typeof editor) {
+      this.monacoEditor = monacoEditor
```
JeanBarriere: please remove this monacoEditor from the constructor and the object. Use the editor sent in the `attach()` method

src/app/store/modules/InfosPane.ts
```diff
@@ -0,0 +1,37 @@
+export interface ISLSError {
+    msg: string,
+    line: number,
+    relativeTop: number
+}
+
+const state = {
```
JeanBarriere: If these states are related to errors and cursorline, better moving the name of this module to something else. It is related to InfosPane on the design but may not always be the case. Also, it's not relevent to what we store inside it, and as it's a data we use only for editorErrors apparently, I'd rename it to something similar to your interface name.

Also, may be unrelated, but what happens in case of a multiple cursor ? It seems here we show the error related to the primary cursor, but what if there is multiple errors

adnrs96: I agree the name isn't great. I thought I would rename it later rather than spend time thinking about a good name which always is a rabbit hole for me

src/editor/plugin-error/index.ts
adnrs96: Cannot do that. This editor is different then the one in object from what I understood from the docs. Talk more in the call about this.

src/app/views/Editor/InfosPane.vue
```diff
@@ -6,17 +6,45 @@
...
       class="text-gray-50 mt-3 sticky top-3"
     >
-      0 Suggestions, 0 Error
+      0 Suggestions, {{ errors.length }} Error
     </s-text>
+    <div
+      v-for="error in errors"
+      :key="`error-brick-${error.line}-${error.msg}`"
+      :style="{
+        marginTop: error.relativeTop + 26 + 'px',
```
JeanBarriere: editor.getDomNode().offsetTop

src/app/views/Editor/InfosPane.vue
```diff
@@ -6,17 +6,45 @@
...
+        marginLeft: cursorLine === error.line? '-2rem': 'unset'
+      }"
+      class="transition-all-fastest"
```
JeanBarriere: https://vuejs.org/v2/guide/transitions.html#List-Entering-Leaving-Transitions

src/app/store/modules/InfosPane.ts
adnrs96: Just to note it here
* Rename to EditorErrors
* Talk to Steve about multiple cursors

adnrs96:
@JeanBarriere @Arinono 
I have updated this PR and I believe its ready for another look. Fixed the e2e test as well which was giving troubles.

adnrs96:
Questions we need to answer:
* What should happen in case of multiple cursors?
* Should time be invested in appearance and disappearance animations. Note: they weren't part of the original acceptance so that can/should be a different story in general.
* What about the code editing panel? I don't know about others but on my laptop (Screen resolution 1366x768) I feel the editor pane to be kinda conjusted especially WRT to gmal label example. I feel moreso since error cards slide animation requires some space on the editor right side so that the cards aren't covering up any code.
* Maybe something more as well, depends upon the feedback?

tests/e2e/editorerrors.e2e.ts
```diff
@@ -0,0 +1,46 @@
...
+  /**
+       * And When I edit the code and SLS reports a error
+       * I see a error card.
+       */
+  it('should display a error card on error', async () => {
+    expect.assertions(2)
+    const context = await page.mainFrame().executionContext()
+    await context.evaluateHandle('monaco.editor.getModels()[0].setValue("http")')
+    // Long wait here since API call involved.
+    await page.waitFor(2500)
```
Arinono: You can have a look to the `.waitForReponse` method that is used for these kind of await.

src/app/views/Editor/InfosPane.vue
```diff
@@ -1,22 +1,138 @@
...
+
+  @Getter('getSLSErrors')
+  private textModelErrors!: ITextModelUriToSLSError
+
+  @Mutation('setTextModelDecorationRendered')
+  private markDecorationRendered!: (textModelUri: string) => void
+
+  private dottedErrors:ITextModelUriToDecorations = {}
+  private highlightDecoration!: IHighlightDecoration | undefined
+  readonly highlightOptions: editor.IModelDecorationOptions = {
```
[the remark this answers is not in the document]
adnrs96: from readonly I understood this is similar to const type and since I don't want the highlightOptions to changeI wanted to use readonly here.
Thoughts on what should be done?

src/app/store/modules/EditorErrors.ts
```diff
@@ -0,0 +1,83 @@
...
+      decorationRendered: false,
+      uri: textModelUri
+    })
+  },
+  setTextModelDecorationRendered: (state: any, textModelUriString: string) => {
+    // Don't need to use Vue.set over here since we don't need reactvity here.
+    state.errors[textModelUriString].decorationRendered = true
+  }
+}
+
```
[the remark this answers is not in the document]
adnrs96: ahh, yes. I kinda forgot to add the typing for state. Will update

src/app/views/Editor/InfosPane.vue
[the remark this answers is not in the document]
Arinono: I usually do `private readonly`, private for the private access (even if once transpiled it doesn't make sense anymore), and readonly for the same purpose as your did (readonly variable)

src/editor/plugin-error/index.ts
```diff
@@ -0,0 +1,81 @@
...
+
+interface ISetModelMarkers {
+    (model: editor.ITextModel, owner: string, markers: editor.IMarkerData[]): void
+}
+
+export class StoryscriptError implements EditorPlugin {
+    private readonly monacoEditor: typeof editor
+    private readonly setModelMarkersOriginal: ISetModelMarkers
+
+    constructor (monacoEditor: typeof editor) {
```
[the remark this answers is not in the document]
adnrs96: This doesn't work. It says `Cannot use namespace 'editor' as a type`
But typeof works fine. Modelled after what is done in `src/editor/core/plugins/monaco.ts`

src/app/store/modules/EditorErrors.ts
```diff
@@ -0,0 +1,83 @@
...
+export interface ISLSErrorState {
+  errors: ITextModelUriToSLSError,
+  cursorLine: number
+}
+
+const state: ISLSErrorState = {
+  errors: {} as ITextModelUriToSLSError,
+  cursorLine: -1
+}
+
```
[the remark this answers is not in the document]
adnrs96: Addressed

src/app/views/Editor/InfosPane.vue
```diff
@@ -1,22 +1,138 @@
...
+  >
     <s-text
+      id="infos-counter"
       p="4"
       weight="medium"
       align="center"
       class="text-gray-50 mt-3 sticky top-3"
     >
-      0 Suggestions, 0 Error
+      0 Suggestions, {{ errors.length }} Error
```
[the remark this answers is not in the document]
adnrs96: Addressed

src/app/views/Editor/InfosPane.vue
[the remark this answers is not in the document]
adnrs96: Addressed

src/app/views/Editor/InfosPane.vue
```diff
@@ -1,22 +1,138 @@
...
+  private cursorLine!: number
+
+  @Getter('getSLSErrors')
+  private textModelErrors!: ITextModelUriToSLSError
+
+  @Mutation('setTextModelDecorationRendered')
+  private markDecorationRendered!: (textModelUri: string) => void
+
+  private dottedErrors:ITextModelUriToDecorations = {}
+  private highlightDecoration!: IHighlightDecoration | undefined
```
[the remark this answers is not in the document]
adnrs96: removed `!` to address

tests/e2e/editorerrors.e2e.ts
adnrs96: Addressed. Thanks for suggesting. BTW actually its `.waitForSelector` :)

src/app/store/modules/EditorErrors.ts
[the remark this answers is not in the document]
adnrs96: Addressed

src/app/views/Editor/InfosPane.vue
```diff
@@ -1,22 +1,137 @@
...
+  >
     <s-text
+      id="infos-counter"
       p="4"
       weight="medium"
       align="center"
       class="text-gray-50 mt-3 sticky top-3"
     >
-      0 Suggestions, 0 Error
+      0 Suggestion, {{ errors.length }} Error{{ errors.length === 0 ? '' : 's' }}
```
Arinono:
```suggestion
      0 Suggestion, {{ errors.length }} Error{{ errors.length > 1 ? 's' : '' }}
```
Sorry I made the mistake earlier 😊

adnrs96: oops and I copy pasted :laughing:

src/editor/core/views/MonacoEditor.vue
```diff
@@ -141,3 +141,14 @@ export default class MonacoEditor extends Vue {
   }
 }
 </script>
+
+<style>
+.monaco-editor .dotted-error {
+    border-bottom-width: 2px;
+    border-bottom-style: solid;
+    box-sizing: border-box !important;
```
JeanBarriere: please, avoid using the `!important` css rule

src/app/types/EditorErrors.ts
```diff
@@ -0,0 +1,24 @@
+import { IRange, Uri } from 'monaco-editor/esm/vs/editor/editor.api'
+
+export interface ISLSError {
+    msg: string,
+    line: number,
+    relativeTop: number,
+    editorTop: number,
+    textModelUri: Uri,
+    range: IRange
+}
```
Arinono: No need to change, but how this file went thought the linter with 4 spaces tabs ?

src/app/views/Editor/InfosPane.vue
```diff
@@ -1,22 +1,103 @@
...
-import { Component, Vue } from 'vue-property-decorator'
+import { Component, Vue, Watch } from 'vue-property-decorator'
+import { Getter, Mutation } from 'vuex-class'
+import { ISLSError, ITextModelUriToSLSError, ITextModelUriToDecorations } from '&/EditorErrors'
+import { editor, Uri } from 'monaco-editor/esm/vs/editor/editor.api'
+import StoryscriptErrorPlugin from '@editor-plugin-error'
 
 @Component({
   name: 'InfosPane'
 })
```
Arinono: No need to change that either, but a reminder that @JeanBarriere made a classic french mistake, `Info` (Information) never takes an `s` in English. This should be `InfoPane`
Same for the filename and stuff.

src/editor/core/views/MonacoEditor.vue
adnrs96: I will remove this. Thanks for pointing out. In fact I would change this thing to 
```
.monaco-editor .dotted-error {
    background: none;
    border-style: dotted;
    padding-bottom: 15px;
    border-bottom-width: 2px;
    border-bottom-color:#EF4E4E;
}
```

src/editor/core/views/MonacoEditor.vue
```diff
@@ -141,3 +141,14 @@ export default class MonacoEditor extends Vue {
...
+    border-style: dotted;
+    border-bottom-color:#EF4E4E;
+    background: none;
```
Arinono: `#EF4E4E` is the tailwind color for `red-50`.  Change your `style` lang to `scss` and use apply to use the tailwind generated classes : `border-b-2 border-dotted border-red-50` and remove `border-bottom-style: solid` since it's not used.

src/editor/core/views/MonacoEditor.vue
Arinono: @adnrs96 @JeanBarriere have a look to my comment below please 👍

src/app/views/Editor/InfosPane.vue
JeanBarriere: Oh yeah, pardon my french :angel:

src/editor/core/views/MonacoEditor.vue
JeanBarriere: :+1:

src/app/views/Editor/InfosPane.vue
Arinono: 🥖 🇫🇷

src/app/types/EditorErrors.ts
adnrs96: well don't know. Like I just assumed linter fixes stuff when I commit.

adnrs96:
@JeanBarriere @Arinono I added a commit fixing the css. Also fixed the 4-space indentation thing @Arinono just for consistency sake.

adnrs96:
@JeanBarriere I have updated this PR with new commits which
- Add support for properly detaching an editor when its destroyed
- Hide the InfosPane (basically destroy it when tab shift happens). Though I don't destroy the error state since code is supposed to be readonly in other tabs and hence when we return to code tab the same errors should still exist.
- Above also fixes the issue of errors from one sample showing up in InfosPane of the other sample when we move from one example to another.
- Also I believe the logic for top margin calculation is fine and with the above changes (specifically destroying the InfosPane on tab change) we should not be seeing the errors with wrong margin tops anymore now.

Can you please take another look?
Thanks!

JeanBarriere:
> * Hide the InfosPane (basically destroy it when tab shift happens). Though I don't destroy the error state since code is supposed to be readonly in other tabs and hence when we return to code tab the same errors should still exist.

I think it's a false assumption. User could go to another story in the meantime and won't have the same nor any errors in the story he'll load. I'd recommend attaching the state to the specific story view and editors, cleaning it on destroy, or making sure it's fresh when we load it

JeanBarriere:
After talk, the assumption is indeed true as the editor instance is destroyed when switching stories

PR #226, opened 2019-12-31
feat(monaco-editor): update monaco-editor version

JeanBarriere:
closes #214 
closes #216 

- [x] regenerate yarn.lock
- [x] fix editor options


src/editor/core/types/editor.ts
```diff
@@ -20,12 +20,12 @@ export const EditorOptions: IEditorConstructionOptions = {
   occurrencesHighlight: false,
   contextmenu: false,
   folding: false,
-  autoIndent: true,
+  autoIndent: 'full',
```
JeanBarriere: I assumed here the indentation was supposed to have all behaviors based from the new options:
`none' | 'keep' | 'brackets' | 'advanced' | 'full`

src/editor/core/types/editor.ts
```diff
@@ -20,12 +20,14 @@ export const EditorOptions: IEditorConstructionOptions = {
...
   cursorBlinking: 'blink',
   cursorWidth: 1,
   lineNumbers: 'off',
   glyphMargin: true,
-  wordBasedSuggestions: false,
+  suggest: {
+    showWords: false
+  },
```
JeanBarriere: `wordBasedSuggestions` has been removed from the list of options. I've replaced it with the `suggest` object having `showWords` to `false`

PR #247, opened 2020-01-08
Feat/remove tips

adnrs96:
### Acceptance

1. Remove all guided tours, their data, tests, etc.
2. Including "Shot Tips" in the Help dropdown

We will be using Intercom exclusively for product tours

Work:
- Remove the Intro component.
- Remove the Tips data store in vuex.
- Remove all occurances of related variables from the above two.
- Remove Tip component which wasn't really being used to display Tips and was sittting dormant. This has been done as a separate commit in case this wasn't what was required so that I can easily drop that last commit.


tests/unit/internal/components/HelpDrop.spec.ts
```diff
@@ -1,18 +1,15 @@
 import { shallowMount, Wrapper, createLocalVue } from '@vue/test-utils'
 import HelpDrop from '@internal/components/HelpDrop.vue'
-import Vuex, { Store } from 'vuex'
-import Tips, { IStoreTips } from '@app/store/modules/Tips'
+import Vuex from 'vuex'
 
 const localVue = createLocalVue()
 localVue.use(Vuex)
```
Arinono: If the TipsStore was the only one needed here, you can also remove the imports an uses of `vuex` and `createLocalVue`

adnrs96: addressed now.

tests/unit/app/views/Layout/index.spec.ts
```diff
@@ -15,9 +14,9 @@ describe('Layout index', () => {
 
   beforeEach(() => {
     store = new Vuex.Store({
-      state: { ...StorePayload.state, ...StoreTips.state },
-      getters: { ...StorePayload.getters, ...StoreTips.getters },
-      mutations: { ...StorePayload.mutations, ...StoreTips.mutations }
+      state: { ...StorePayload.state },
+      getters: { ...StorePayload.getters },
+      mutations: { ...StorePayload.mutations }
     })
```
JeanBarriere: this can be moved to a simple new `Vuex.Store(StorePayload)` then

adnrs96: I didn't do that because the way it is written right now allows or actually creates a copy of the configuration present inside StorePayload
I faced some issues in my previous SLS-Errors PR while writing tests where if I did, `store = new Vuex.Store(Foobar)` inside `beforeEach` changes from one test to the store still persisted over to the other.
When we do things this way, everytime a new config is generated and I believe its a better way to do things.

adnrs96: Though if you feel that we should change it to `Vuex.Store(StorePayload)` I am happy to do that :)

JeanBarriere: `Vuex.Store({ ...StorePayload }) if you want a cleaner copy I guess

adnrs96: Okay addressed using @JeanBarriere last comment. Thanks for suggesting it :)

PR #250, opened 2020-01-09
feat(Studio): add whitelisted banner according to design.

adnrs96:
### Acceptance

**When** I navigate to any Story (new or example)
**Then** the whitelist warning appears according to the designs

Work -

This banner is being shown all the time right now, but
this would actually be only shown to users after they have
logged into the studio. This commit is more of just
introducing the banner to codebase. We will add conditional
display in later commits.


adnrs96:
Comment/Note:
I took a look at the https://github.com/storyscript/components/blob/master/src/assets/styles/custom/_backgrounds.scss to get an idea about how to add the gradient CSS.

I wasn't able to use the `@include linear-gradient` because for some reason SASS loader kept on saying the mixin doesn't exist but it clearly seemed to exist in the `_gradient.scss` stylesheet.

Arinono:
> Comment/Note:
> I took a look at the https://github.com/storyscript/components/blob/master/src/assets/styles/custom/_backgrounds.scss to get an idea about how to add the gradient CSS.
> 
> I wasn't able to use the `@include linear-gradient` because for some reason SASS loader kept on saying the mixin doesn't exist but it clearly seemed to exist in the `_gradient.scss` stylesheet.

You can take what's on figma and make it work with tailwind. This is what I did for `www.story.io`:
```
.rainbow-bar {
  @apply w-full;
  height: 5px;
  background: linear-gradient(177.72deg, #9650FD 4.79%, #F24323 19.06%, #FF9288 33.61%, #1D78E2 51.03%, #1AB3FA 67.64%, #13BEB2 86.32%, #0DC775 102.6%);
}
```

JeanBarriere:
the mixin `linear-gradient` has been defined in the mixins folder here: https://github.com/storyscript/components/blob/master/src/assets/styles/mixins/_gradients.scss

The reason behind it is to have this mixin defined, which was almost like it in the compass dependencies but as compass is now deprecated and no longer maintained, it is not a good idea to add this dependency. Although you can still reuse the mixin here by importing it

adnrs96:
Hey @JeanBarriere @Arinono 
I have updated this PR with:
- show banner only when logged in.
- tried to use what @Arinono suggested for CSS.
- added e2e test in login.e2e.ts and tested it against a deployment and things went smoothly.

adnrs96:
RE:
> the mixin linear-gradient has been defined in the mixins folder here:
I tried to use that function. it requires the syntax `background: @include linear-gradient(direction, ...[color stops])` to best of my knowledge and I see the following error
```
Failed to compile.

./src/app/views/Layout/index.vue?vue&type=style&index=0&lang=scss& (./node_modules/css-loader/dist/cjs.js??ref--8-oneOf-1-1!./node_modules/vue-loader/lib/loaders/stylePostLoader.js!./node_modules/postcss-loader/src??ref--8-oneOf-1-2!./node_modules/sass-loader/dist/cjs.js??ref--8-oneOf-1-3!./node_modules/cache-loader/dist/cjs.js??ref--0-0!./node_modules/vue-loader/lib??vue-loader-options!./src/app/views/Layout/index.vue?vue&type=style&index=0&lang=scss&)
Module build failed (from ./node_modules/sass-loader/dist/cjs.js):
SassError: Expected expression.
    ╷
187 │     background: @include linear-gradient(
    │                 ^
    ╵
  /home/adnrs96/studio/src/app/views/Layout/index.vue 187:17  root stylesheet
```

Therefore I prefered what @Arinono  suggested and went with it. I don't know why use @include doesn't work.

PR #270, opened 2020-01-14
Feat/gh e2e

adnrs96:
work

- Add e2e test for login workflow.

this test doesn't pass yet since the tests won't work in spun up environment.

If you have an environment you need to replace TEST_URL with the deployment environment string to be able to run the test against it


Arinono:
> this test doesn't pass yet since the tests won't work in spun up environment.
If you have an environment you need to replace TEST_URL with the deployment environment string to be able to run the test against it

I tried with an environment made by Will concourse pipeline and it works fine. I'm disabling this test for now until we figure out to run these E2E tests outside Netlify/GH Actions

PR #274, opened 2020-01-15
feat(logout): add logout process

JeanBarriere:
- [x] add dropdown on the navbar's user avatar
- [x] add logout action
- [x] add tests

---
**changes for docker env**
- need to add LOGOUT_URL env variable

---
### Acceptance
**Given** I am logged in to the studio
**When** I click on my avatar
**Then** I see a dropdown with a logout option

**And when** I click the logout option
**Then** my page is refreshed and I see the sign up button in the top right where my avatar was


src/app/views/Layout/Navbar.vue
```diff
@@ -66,6 +67,7 @@ import { AccentColorsEnum } from '@storyscript/vue'
     SHelp,
     SRegisterInterest,
     SAvatar,
```
adnrs96: If SAvatar isn't used anymore should we remove this import?

PR #286, opened 2020-01-17
feat(docs): add documentation in the library pane

Arinono:
Feedback:
- [x] Remove hash from URL, and use vuex state
- [x] Remove coma in method params
- [x] Add new env to all `.env.*` and Docker image
- [x] Tests

---

The motivation here is to create an architecture and a set of really simple components so that everyone can write documentation without writing a single line of Javascript, and as little as possible of regular HTML.

[Here is](https://github.com/storyscript/studio/blob/d8061a6f5ab3ff61e363c364e844b53d1a872494/src/app/views/Library/Docs/Types/String/index.vue) an example of what I have in mind for the documentation (in a single file, no component declaration)
```html
<template>
  <docs-container
    :show="['#docs-Types', '#docs-Types-String']"
    to="#docs-Types-String"
  >
    <docs-row class="px-3 mt-2 mb-1">
      <docs-icon
        icon="double-quotes"
        class="mr-2"
      />
      <docs-title>String</docs-title>
    </docs-row>
    <docs-desc>
      <docs-text class="mb-4">
        A single block of text used to store text in any size.
      </docs-text>
      <docs-container :show="['#docs-Types']">
        <docs-code class="mb-2">
          "hello world"
        </docs-code>
        <docs-code>
          "hello {name}"
        </docs-code>
      </docs-container>
    </docs-desc>
  </docs-container>
</template>
```

And here is the result:
Screenshot 2020-01-17 at 20 04 43


JeanBarriere:
Having a JSON Blob wouldn't be better here ? It would avoid misuse of your component structure.
Thoughts ?

src/app/services/DocsFetcher.ts
```diff
@@ -0,0 +1,24 @@
+import yaml from 'js-yaml'
+
+export default class DocsFetcher {
+  private _url: string
+
+  constructor (url: string) {
+    if (!url || url.length <= 0) {
```
JeanBarriere: I think `!url` will also handle the case where `url = ''`

src/app/store/index.ts
```diff
@@ -12,6 +12,6 @@ export default new Vuex.Store({
   strict: env('NODE_ENV') !== 'production',
   plugins: [createPersistedState({
     key: `studio-v${version.split('.')[0]}`, // keeping storage away from potential breaking changes
-    paths: ['LibraryPane', 'Panes']
+    paths: ['LibraryPane', 'Panes', 'Docs']
```
JeanBarriere: Why persisting the docs state ?

src/app/views/Layout/index.vue
```diff
@@ -37,9 +35,15 @@
             v-if="isSelected('Story')"
             class="inline-flex relative"
           >
-            <s-panes class="z-10">
-              <s-pane :pane="{ name: 'Library', icon: 'book-open' }">
-                <s-library />
+            <s-panes
+              class="pt-3 z-10 transition-all-slow-linear"
```
JeanBarriere: please remove these classes, maybe except `z-10`

src/app/views/Layout/index.vue
```diff
@@ -37,9 +35,15 @@
...
+              side="left"
```
JeanBarriere: side is no longer required

src/app/views/Layout/index.vue
```diff
@@ -37,9 +35,15 @@
...
+            >
+              <s-pane
+                :pane="{ name: 'Library', icon: 'book-open' }"
+                class="w-full"
```
JeanBarriere: this class should not be needed anymore

src/app/store/index.ts
Arinono: Because it's annoying right when you refresh to lose the docs you had. Later we will be able to remove that since the docs will be displayed depending on the editor context.

src/app/services/DocsFetcher.ts
Arinono: I will check 👍

Arinono: Addressed

src/app/views/Layout/index.vue
Arinono: Addressed

src/app/views/Layout/index.vue
Arinono: Addressed

src/app/views/Layout/index.vue
Arinono: Addressed

tailwind.config.js
```diff
@@ -175,6 +175,7 @@ module.exports = {
       wider: '.02em' // 2%
     },
     fontSize: {
+      '2xs': '.625rem', // 10px
```
JeanBarriere:
```suggestion
      'xxs': '.625rem', // 10px
```

more consistent nah ?

src/internal/components/docs/Text.vue
```diff
@@ -0,0 +1,14 @@
...
+    weight="regular"
+    class="text-gray-60"
+    v-text="$slots.default[0].text"
+  />
+</template>
+
+<script lang="ts">
+import { Component, Vue } from 'vue-property-decorator'
+
+@Component({ name: 'DocsText' }) export default class DocsText extends Vue { }
```
JeanBarriere: one line now ? :cry:

src/internal/components/docs/Icon.vue
```diff
@@ -0,0 +1,15 @@
...
+    :icon="icon"
+    class="text-gray-80"
+  />
+</template>
+
+<script lang="ts">
+import { Component, Vue, Prop } from 'vue-property-decorator'
+
+@Component({ name: 'DocsIcon' }) export default class DocsIcon extends Vue {
+  @Prop({ type: String }) readonly icon!: string
```
JeanBarriere: Add required: true or default value here

src/internal/components/docs/Code.vue
```diff
@@ -0,0 +1,16 @@
+<template>
+  <div class="bg-gray-10 rounded-md border-gray-20 border px-2 py-1 w-fit-content">
+    <s-text
+      p="4"
+      weight="medium"
+      class="font-monospace text-black w-fit-content"
+      v-text="$slots.default[0].text"
```
JeanBarriere: what happens if there is no slot provided ?

src/internal/components/docs/Card.vue
```diff
@@ -0,0 +1,117 @@
...
+  @Prop({ type: String, required: true }) readonly title!: string
+  @Prop({ type: String, default: undefined }) readonly returnType!: string | undefined
+  @Prop({ type: String, default: '' }) readonly description!: string
+  @Prop({ type: String, default: undefined }) readonly to!: string
+  @Prop({ type: Array, default: () => ([]) }) readonly more!: []
+  @Prop({ type: String, default: undefined }) readonly return!: string
+  @Prop({ type: Object, default: undefined })
+  readonly icon!: { path: string, external?: boolean } | undefined
+
+  private docsRoot: string = process.env.VUE_APP_DOCS_ROOT || ''
```
JeanBarriere: should we have a default to /docs here ? Or something more useable than root = '' ?

src/internal/components/Tag.vue
```diff
@@ -0,0 +1,40 @@
+<template>
+  <div
+    class="rounded-xs flex items-center justify-center px-1"
+    :class="[bgColor]"
+  >
+    <div
+      class="font-monospace font-semibold text-2xs"
```
JeanBarriere: reference to the `2xs` to `xxs`

src/internal/components/Tag.vue
```diff
@@ -0,0 +1,40 @@
+<template>
+  <div
+    class="rounded-xs flex items-center justify-center px-1"
+    :class="[bgColor]"
```
JeanBarriere: no need of these brackets here

src/internal/components/Tag.vue
```diff
@@ -0,0 +1,40 @@
...
+  >
+    <div
+      class="font-monospace font-semibold text-2xs"
+      :class="[
+        textColor
+      ]"
```
JeanBarriere: same here, just having :class="textColor" will work

src/internal/components/Breadcrumb.vue
```diff
@@ -0,0 +1,62 @@
...
+import { Component, Vue, Prop } from 'vue-property-decorator'
+
+@Component({
+  name: 'SBreadcrumb'
+})
+export default class SBreadcrumb extends Vue {
+  @Prop({
+    type: Array,
+    required: true
+  }) readonly paths!: Array<string>
```
JeanBarriere: Why are these multilines now ? :cry:

src/app/views/Docs/MainMenu.vue
```diff
@@ -0,0 +1,77 @@
...
+  onMenuChange () {
+    Object.keys(this.menu).forEach((key: any) => {
+      this.$set(this.menus, key, this.menu[key])
+    })
+  }
+
+  private goTo (key: string): void {
+    const menu: any = this.menus[key]
+    const hash: string = menu.to
+    if (menu.available !== false && this.current !== hash) {
```
JeanBarriere: !!menu.available ?

tailwind.config.js
Arinono:
```
fontSize: {
      '2xs': '.625rem', // 10px
      xs: '.75rem', // 12px
      sm: '.875rem', // 14px
      base: '1rem', // 16px
      lg: '1.125rem', // 18px
      xl: '1.25rem', // 20px
      '2xl': '1.5rem', // 24px
      '3xl': '1.875rem', // 30px
      '4xl': '2.25rem', // 36px
      '5xl': '2.5rem', // 40px
      '6xl': '3.75rem' // 60px
    }
```

This is what I call consistency 🤔

src/internal/components/docs/Text.vue
Arinono: Yeah my bad, I'll fix all of these 🙇

src/internal/components/docs/Code.vue
Arinono: Nothing is displayed ? (+ the YAML parsing will break first)

```
...
more:
      - code: # Here you cannot leave it empty, so an empty string (at least) has to be provided
      - code: '"hello {name}"'
...
```

But you right I can secure this to avoid logging errors in the console

src/internal/components/Breadcrumb.vue
Arinono: Probably because there was a validator (so line was too long) and I forgot to put it back on 1 line.

PR #292, opened 2020-01-20
feat(monaco): use Monarch for syntax-highlighting + add RichLanguageConfiguration

wilzbach:
(supersedes https://github.com/storyscript/studio/pull/283 by using the new https://github.com/storyscript/monaco-storyscript-language package - https://www.npmjs.com/package/@storyscript/monaco-language)

Fixes https://github.com/storyscript/sls/issues/93
Closes #283 

This is a rewrite of Storyscript's grammar into the Monarch format. Moreover, this PR adds an initial RichLanguageConfiguration which allows e.g. automatically adding (and detecting) braces:

Peek 2020-01-17 11-13

### What is Monarch?

> This library [Monarch] allows you to specify an efficient syntax highlighter, using a declarative lexical specification (written as a JSON value). The specification is expressive enough to specify sophisticated highlighters with complex state transitions, dynamic brace matching, auto-completion, other language embeddings, etc. as shown in the 'advanced' topic sections of this document

### Why not use the monaco-ace-tokenizer?

1) Monaco's own syntax parser [Monarch](https://microsoft.github.io/monaco-editor/monarch.html) is getting more stable and is actively improved/developed (see e.g. https://github.com/microsoft/monaco-languages/commits/master)
2) Monarch is more powerful (it e.g. "Über-Advanced" which allows syntax-highlighting for embedded languages). This will be interesting for upcoming language features
3) Monarch output is easier to understand and maintain (it's less than 200 lines) 
4) No Iro middle-man (the current grammar was in Iro which the last update being from 2017 and thus locking us into Ace)
5) Monarch allows to write nice automated tests (see below) 
6) Monaco-ace-tokenizer was just a temporary solution. The library says so itself:

> This library is relevant only till monarch definitions of all the remaining languages are added directly in monaco-editor itself. Untill then, it can be used if you do not want to use web assembly.

https://github.com/brijeshb42/monaco-ace-tokenizer

(7) Using the native Monarch parser should be faster than the Ace token emulation, but I haven't tested or measured this.)

### I) Test suite

As mentioned Monarch allows to write tests and I have done so in my fork of monaco-languages:

https://github.com/microsoft/monaco-languages/compare/master...wilzbach:storyscript

We have the following options:
1) Try to make a PR to the official monaco-languages (limits our abilities to make quick language changes)
2) Make my fork of monaco-languages Storyscript-official (i.e. move into the Storyscript organization) and
  2.a) include as submodule
  2.b) publish on npm and include as npm dependency
3) Include inline as in this PR

The current option (3) is only intended to be a show-case solution as we don't have the ability to run tests easily here. I would recommend 2.a or 2.b

### II) Colorscheme

Currently the IDLE theme from https://github.com/brijeshb42/monaco-themes/blob/master/themes/IDLE.json is used. However, it's based on Ace themes and thus it uses a few non-standard class names. However, as we know have the chance to put everything into one place we have two options:

a) copy over the IDLE theme and customize it further (i.e. we could use a different color for `function`)
Advantage: further theme customization
Disadvantage: no longer compatible with the existing Ace themes (if we only plan to support 1-3 themes this should be no problem)
b) leave as is with the two non-standard Monarch CSS classes (see e.g. `support.function`).

Note: that I mostly tried to use the standard Monarch CSS classes e.g. `number` instead of `constant.numeric` or `string` instead of `storage.type.string` which is why the output now looks slightly different. As can be seen below the in the source code, we have full control over these classes (b), but as they are used also for all official `monaco/monaco-languages` I would prefer using them and instead adapt the theme (a).

Current:

image

New:

image


package.json
```diff
@@ -18,6 +18,7 @@
     "test:integration-headless": "TEST_MODE=integration vue-cli-service test:unit --runInBand"
   },
   "dependencies": {
+    "@storyscript/monaco-language": "^0.0.1",
```
JeanBarriere: can you pin this dependency ? Otherwise, we'll get a renovate PR generated that can be avoided. Thanks :)

wilzbach: Sure. Updated.

PR #296, opened 2020-01-20
Feat/diagram panel

adnrs96:
### Acceptance

**When** I am viewing any Story
**Then** I see a panel-toggle for Diagram according to the designs

**When** I click on the Diagram panel-toggle
**Then** I the Diagram panel slides in (like the Library but on the right side)
**And** I see the preview state of the Diagram (like in the Diagram tab in Events/Services tab)

Work

- Add diagram Panel.
- Add unit tests.
- Add e2e tests.


adnrs96:
I am yet to start using the icon given in design.
I have a commit ready for it but couldn't create a PR to components Lib due to issue with permissions :(

adnrs96:
@JeanBarriere @Arinono This is ready for a review I think.
Shortcomings I can think of
- Overlaps with infospane.
- Overlaps with editor in viewing mode.

PR #298, opened 2020-01-21
improvement(sentry): enhance sentry with release, environment and commit infos

wilzbach:
I recently added more Sentry information to SLS and this was very helpful for me. Thus, I thought this might help the Studio team as well.

See e.g. https://docs.sentry.io/cli/releases or https://blog.sentry.io/2017/05/01/release-commits for details.


.github/workflows/sentry.yml
```diff
@@ -0,0 +1,27 @@
...
+      - uses: actions/checkout@v1
+
+      - name: Install sentry
+        run: curl -sL https://sentry.io/get-cli/ | bash
+
+      - name: Link commits with sentry releases
+        env:
+          SENTRY_AUTH_TOKEN: ${{ secrets.SENTRY_AUTH_TOKEN }}
+          SENTRY_ORG: ${{ secrets.SENTRY_ORG }}
+          SENTRY_PROJECT: ${{ secrets.SENTRY_PROJECT }}
```
wilzbach: I have already configured these variables here.

.github/workflows/sentry.yml
```diff
@@ -0,0 +1,27 @@
...
+        run: |
+          sentry-cli releases new ${GITHUB_SHA}
+          sentry-cli releases set-commits --auto ${GITHUB_SHA}
+          sentry-cli releases finalize ${GITHUB_SHA}
```
wilzbach: This is the most interesting part. There are basically two options:
- use commit shas as releases
- use releases
See https://docs.sentry.io/cli/releases/#creating-releases for details.

I wasn't sure what works best for you, but considering that you currently don't expose the current commit sha to the app, you might want to only do this for relases tags.

src/internal/plugins/sentry.ts
```diff
@@ -1,10 +1,13 @@
 import Vue from 'vue'
 import * as Sentry from '@sentry/browser'
 import * as Integrations from '@sentry/integrations'
+import { version } from '@/../package.json'
 
 if (process.env.NODE_ENV === 'production') {
   Sentry.init({
     dsn: process.env.VUE_APP_SENTRY_DSN || '',
+    release: version,
+    environment: location.host,
```
wilzbach: I couldn't find a better distinguisher between the master and dev environments.

src/internal/plugins/sentry.ts
```diff
@@ -1,10 +1,13 @@
 import Vue from 'vue'
 import * as Sentry from '@sentry/browser'
 import * as Integrations from '@sentry/integrations'
+import { version } from '@/../package.json'
 
 if (process.env.NODE_ENV === 'production') {
   Sentry.init({
     dsn: process.env.VUE_APP_SENTRY_DSN || '',
+    release: version,
```
wilzbach: This is very helpful as it allows one to know in what deployed version the error occurred. Hence, even more detailed information (e.g. commit shas) can help more.

wilzbach:
> Is there a value in using (or making if we don't trust 3rd-party sources) Sentry specific GH Actions ?

There definitely is, but all the third-party actions look rather new and small (<10 stars) and the official GH app is limited to a single user only in the free plan:

image

Sending this single request isn't very complicated and at least this way we have full control. Though if we hit a problem, we could always try to see how far the Sentry Free/Business plan takes us ...

PR #304, opened 2020-01-22
Feat/suggesting

adnrs96:
### Acceptance
**When** I click the Viewing toggle
**Then** I see suggestion mode as an option

**When** Suggestion mode is toggled on
**Then** The all write-features (editing, adding sections, etc) is disabled (since this is a read-only)
**And** I see a fake suggestion card according to the designs that is static on the page, regardless of where you scroll.

Work
- Add Suggesting Option
- Add Suggesting Coming soon card
- Add unit tests
- Add e2e tests
- Some other work to make things work :wink:


tailwind.config.js
```diff
@@ -153,6 +153,7 @@ module.exports = {
         min: '.125rem',
         1: '.25rem',
         3: '.75rem',
+        13: '3.125rem',
```
JeanBarriere: Just a reminder here that we need to get away with this static top value in the future

adnrs96: Do you mean start doing something like we have for negative top's?
Basically `-top-5` would just work?

PR #311, opened 2020-01-23
feat(Topbar): select viewing mode based on user logged in/out.

adnrs96:
If user is logged in, story is in Editing mode by default
If user is logged out, story is in Viewing mode by default


src/app/views/Layout/Topbar.vue
```diff
@@ -186,10 +186,18 @@ export default class STopbar extends Vue {
     }
   }
 
+  private async updateViewingMode () {
+    const result = await this.$apollo.query({ query: require('@graphql/queries/viewer.gql') }).catch(() => ({ data: {} }))
+    if (result && result.data && result.data.viewer) {
```
Arinono:
```suggestion
    if (result?.data?.viewer) {
```

It's not important, but you can use [this syntax](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#optional-chaining) now (even if the Vetur is not yet up to date with this version of TS)

adnrs96: Will keep this in mind. This is cool syntax IMO. Thanks for letting me know :)

PR #320, opened 2020-01-24
feat(sls): indentation as a custom LSP feature

wilzbach:
Supersedes https://github.com/storyscript/studio/pull/207
Fixes https://github.com/storyscript/sls/issues/127

TODO:
- [ ] integration tests


vue.config.js
```diff
@@ -33,6 +35,16 @@ module.exports = {
...
         test: /\.story$/,
         use: 'raw-loader'
+      }, {
+        // include ES6 modules in transpilation to ES5
+        test: /\.js$/,
+        loader: 'babel-loader',
+        exclude: {
+          test: path.resolve(__dirname, 'node_modules'),
+          exclude: [
+            path.resolve(__dirname, 'node_modules/vscode-languageclient') // uses ES2017
```
wilzbach: This was rather nasty to figure out, but it turns out that the vscode-languageclient module is compiled down to es2017.
https://github.com/microsoft/vscode-languageserver-node/blob/master/client/tsconfig.json

Apart from being problematic for browser compatibility, this is very important as otherwise we can't extend classes from this module (extending es6 classes from transpiled es5 doesn't work). One gets a nice error like: "class constructors must be invoked with 'new'".

It's a bit curious that this wasn't detected before, but I guess all devs here are just using latest FF|Chrome.

src/editor/plugin-sls/indent.ts
```diff
@@ -0,0 +1,150 @@
...
+            const p = client.sendRequest(IndentationRequest.method, params) as Promise<Indentation>
+            return p.then(
+              (value:Indentation) => {
+                const textEdits = this.sls.p2m.asTextEdits(value.textEdits)
+                model.pushEditOperations(ed.getSelections()!, textEdits, () => {
+                  return null
+                })
+                // monaco selects the editOperation by default. Reset it to the current cursor
+                const sel = ed.getSelection()!
+                ed.setSelection(new monaco.Selection(sel.endLineNumber, sel.endColumn, sel.endLineNumber, sel.endColumn))
```
wilzbach: I'm not sure why monaco does this, but otherwise the newly inserted text gets selected:

image

In theory [pushEditOperations](https://microsoft.github.io/monaco-editor/api/interfaces/monaco.editor.itextmodel.html#pusheditoperations) should allow one to yield the new position of the cursor.
Though as the end position is unknown, this wouldn't be easier.
One could of course compute the new cursor location manually and return it in the `ICursorStateComputer`, but I don't think this would be less complex.

src/editor/plugin-sls/indent.ts
```diff
@@ -0,0 +1,150 @@
...
+          id: 'enter-indent',
+          label: 'Enter Indent Overload',
+          keybindings: [KeyCode.Enter],
+          keybindingContext: undefined, // don't show in context menu
+          run: (ed: IStandaloneCodeEditor): Promise<void> => {
+            const client = this._client
+            const model = ed.getModel()!
+            const params : IndentationParams = this.sls.m2p.asTextDocumentPositionParams(model, ed.getPosition()!)
+            const p = client.sendRequest(IndentationRequest.method, params) as Promise<Indentation>
+            return p.then(
```
wilzbach: TODO: add a timeout and fallback to a manual indent after e.g. 200ms

src/editor/plugin-sls/indent.ts
```diff
@@ -0,0 +1,182 @@
...
+            const p = client.sendRequest(IndentationRequest.method, params)
+            let timeout : NodeJS.Timeout
+            return (Promise.race([
+              p,
+              // add a default timeout to the LSP request
+              new Promise((resolve, reject) => {
+                timeout = setTimeout(() => {
+                  reject(new IndentationTimeoutError())
+                }, this.requestTimeout)
+              })
```
wilzbach: Might be worth generalizing this if this is a more common pattern.

PR #326, opened 2020-01-27
feat(new pane): add new pane design and behaviors

JeanBarriere:
- [x] remove conditional panes in store
- [x] ensure library is first pane to show
- [x] improve panes toggling
- [x] remove open/closing modes
- [x] redesign diagram/toc panes
- [x] redesign pane toggler component
- [x] unit tests

---

### Acceptance

**When** I navigate to a Story
**Then** I see the new panel design

This new panel design is implemented in light of horizontal space being limited causing overlaps with cards (errors, suggestions, etc). Also, no panels work together in a way requiring them to be open together at any given moment.

**When** I toggle Viewing mode
**Then** the Table of Contents toggles in.

**When** I toggle Editing or Suggesting mode
**Then** the Library toggles in.


tailwind.config.js
```diff
@@ -135,11 +135,13 @@ module.exports = {
...
       spacing: {
         initial: 'initial',
+        inherit: 'inherit',
         'fit-content': 'fit-content',
         'max-content': 'max-content',
         '2full': '200%',
         14: '3.5rem', // 56px
         22: '5.5rem', // 88px
+        75: '18.75rem', // 300px
         76: '19rem' // 304px
```
Arinono: Is the difference between 300px and 304px really making a big change ? Maybe we can use only one, like a multiple of 4, as we always did 🌝

tailwind.config.js
```diff
@@ -157,9 +159,15 @@ module.exports = {
...
+      minWidth: {
+        inherit: 'inherit'
+      },
+      minHeight: {
+        inherit: 'inherit'
+      },
       maxWidth: {
         xxs: '10rem',
-        pane: '19rem'
+        pane: '18.75rem'
```
Arinono: If you want to stick with exactly 300px, that's fine, but we should remove the `w-76` then.

tests/unit/app/views/DiagramPanel/index.spec.ts
```diff
@@ -1,78 +1,22 @@
...
-        vm.toggled = false
-        vm.togglePanel()
-        expect(vm.toggled).toBeTruthy()
-        await new Promise(resolve => setTimeout(resolve, 400))
-        expect(vm.toggled).toBeFalsy()
-      })
+  describe('should mount', () => {
+    it('should have html', () => {
+      expect(view.html()).toBeTruthy()
     })
```
Arinono: The `view.destroy()` is missing (missed the last time)

tailwind.config.js
```diff
@@ -157,9 +159,15 @@ module.exports = {
...
       }
```
Arinono: We should keep only one of the two `w-75` `w-76`, they are really close and we don't need both. Maybe we should keep the `w-76` as it's a multiple of 4, and we already agreed on that previously (and I don't think the difference is that big).

tests/unit/app/views/DiagramPanel/index.spec.ts
```diff
@@ -1,78 +1,22 @@
...
-      it('should set toggled to true for 350ms', async () => {
-        vm.toggled = false
-        vm.togglePanel()
-        expect(vm.toggled).toBeTruthy()
-        await new Promise(resolve => setTimeout(resolve, 400))
-        expect(vm.toggled).toBeFalsy()
-      })
+  describe('should mount', () => {
+    it('should have html', () => {
+      expect(view.html()).toBeTruthy()
```
Arinono: WEE WOO 🚨 
There no `view.destroy()` we missed it the last time.

tailwind.config.js
JeanBarriere: Yeah, I just followed design here. I can put back to 76

tests/unit/app/views/DiagramPanel/index.spec.ts
JeanBarriere: :astonished:

JeanBarriere:
@Arinono review addressed

PR #328, opened 2020-01-27
Add skip test helpers

williammartin:
On my local machine I can now get tests actually green:

image

Not sure if I set these env vars correctly in the workflow but we'll see.

Also this is rebased on the TLS PR for now.


tests/e2e/register-interest.e2e.ts
```diff
@@ -65,7 +65,7 @@ describe('Logs', () => {
    */
   it('should close the modal and clear the inputs when submitted', async () => {
     expect.assertions(1)
-    if (!page.url().includes(HOST)) {
+    if (!page.url().includes(TEST_URL)) {
```
Arinono: Why not using the `withNetlify` here ?

Arinono:
Also we can probably add the `Closes #323` on this one and flag the other one as not mergeable. The other PR without this addition is kinda useless :thinking:

tests/e2e/register-interest.e2e.ts
williammartin: Oversight. I didn't know where all the tests were.

williammartin:
Not sure why you say that. The ability to support TLS and the ability to skip unrunnable tests seem separate? Doesn't really matter to me though.

Arinono:
> Not sure why you say that. The ability to support TLS and the ability to skip unrunnable tests seem separate? Doesn't really matter to me though.

It's just because you took the changes of #323 on your branch, so we can just skip one PR, and merge both

PR #332, opened 2020-01-28
feat(AskForHelp): add raise hand for help.

adnrs96:
### Acceptance

> Guests and logged-in users can interact with this preview feature.

**When** I highlight code
**Then** The raise your hand button will appear according to the designs with a subtle, slide in animation 

**When** I click away (no more code is highlighted)
**Then** the raise your hand button slides out

**When** I click the raise your hand button
**Then** a popup, according to the designs.

**If** there is no content in the textarea
**Then** The "Ask for help" button is disabled

**When** I click the button "Ask for help"
**Then** A new Intercom window opens with their text as the message they sent in the chat window
**And** the popup is closed

Work

- [x] Add raise hand for help
- [x] feat(Card): add card component and make raise hand use new card component.
- [x] Add send message to intercom when asking for help
- [x] Add unit tests
- [ ] chore(Errors): use new card component for errors as well


src/app/store/modules/AskForHelp.ts
```diff
@@ -0,0 +1,31 @@
...
+  show: boolean
+}
+
+const state: IAskForHelpState = {
+  ask: {} as IAskForHelp,
+  show: false
+}
+
+const getters = {
+  showAskForHelp: (state: IAskForHelpState) => state.show,
```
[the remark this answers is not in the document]
adnrs96: oh, man! I had this rename in mind before (maybe this morning). I thought I would do it at the end before creating PR as a separate commit.

I will do it now :)
Thanks for pointing it out.

src/app/views/Editor/AskForHelp.vue
```diff
@@ -0,0 +1,196 @@
+<template>
+  <div
+    id="ask-for-help"
+    class="z-20"
+  >
+    <s-card
+      v-show="showAskForHelp"
+      :type="type"
+      :style="{
+        marginTop: `${pos.relativeTop + pos.editorTop - marginTopAdjustment}px`,
```
JeanBarriere:
```suggestion
        top: `${pos.relativeTop + pos.editorTop - marginTopAdjustment}px`,
```

replace by top so we can click on the existing errors and interract with them, marginTopAdjustement may not be needed then

src/editor/plugin-error/index.ts
```diff
@@ -102,14 +102,36 @@ export class StoryscriptError implements EditorPlugin {
...
+            relativeTop: editor.getTopForLineNumber(sln),
+            editorTop: editorTop,
+            textModelUri: textModel?.uri,
+            selection: e.selection
+          })
+          store.commit('setShowAskForHelp', true)
+        } else if (e.selection.isEmpty()) {
+          store.commit('setShowAskForHelp', false)
+        }
+      })
```
JeanBarriere: error and asking for help are not the same features. This is weird and will be difficult to maintain/find in the future. Can you move it somewhere else (create a new plugin) or rename the plugin maybe ?

src/internal/components/Card.vue
```diff
@@ -0,0 +1,51 @@
+<template>
+  <transition name="slide-fade">
+    <div
+      class="flex items-center w-fit-content"
+      :class="[
+        {'p-1/2 border-card-colorful rounded-full': type === 'raiseHand'},
```
JeanBarriere: a Card component having a `raiseHand` type is weird, it seems like it's handling states for the askForHelp component, which does not seem like a good behavior

adnrs96: It isn't handling states IMO. type is just a prop based on which you render different card props.
Card is a component which receives props and does something with those props which change its look.
I am not sure how would we do a component for cards otherwise if we did not do a prop such as type and change looks based on its value.

Maybe you are suggesting pull out the value 'raiseHand' somewhere else and import them here?

JeanBarriere: No, I'm indeed referring to the cardType being `raiseHand` seems a bit too focused on this specific feature and is weird for a component just named `card` which seems highly abstract

src/editor/plugin-error/index.ts
adnrs96: Doing a separate plugin might make sense though I didn't do that because I am kinda not sure how these features are going to play out. And it seems that from monaco's point of view, they aren't really that different things because they all need to use some events from monaco editor.

We setup those editor events to calls functions.

I feel close on the horizon, there will be a need to do some stuff on same event but different for features. Now if we separated the plugins for each feature we will end up in trouble.

src/internal/components/Card.vue
adnrs96: created https://www.pivotaltracker.com/story/show/171004028 for this.

PR #343, opened 2020-01-30
feat: disable guest access

Arinono:
https://www.pivotaltracker.com/story/show/170876079
https://studio.sarphati.storyscript-ci.com/


src/internal/plugins/vue-apollo.ts
```diff
@@ -9,6 +9,7 @@ if (!process || env('NODE_ENV') !== 'test') {
   Vue.use(VueApollo)
 }
 
+let _client: ApolloClient<any>
```
JeanBarriere: just a note here, changing it to Client or apolloClient would be a better name, but that's fine

src/app/guards/LoginGuard.ts
```diff
@@ -0,0 +1,11 @@
+import GQLUserService from '@/app/services/GQLUser'
+
+export default class LoginGuard {
+  constructor (private userService: GQLUserService) { }
+
+  public async beforeEnter (next: Function) {
+    if (!await this.userService.isAuthenticated().catch(() => false)) {
+      window.location.replace('https://storyscript.com')
+    } else next()
+  }
```
JeanBarriere: I understand the logic, this will still be called on every page change, which is maybe overkill.
A more optimized solution would be adding an interceptor to the `apolloClient`, and redirect to `storyscript.com` if we get a 401 on any query.

But this works :ok_hand:

JeanBarriere:
Also, WhitelistBanner is going to be duplicate with my change. I'm removing it from my side

PR #348, opened 2020-01-31
Feat/library pane ui

JeanBarriere:
- [x] update icons
- [x] update library mainMenu component design
- [x] update breadcrumbs component design


src/app/views/Docs/index.vue
```diff
@@ -63,6 +62,7 @@ export default class SDocs extends Vue {
 
   @Watch('current', { immediate: true, deep: true })
   private async onRouteChange () {
+    console.log('current = ', this.current)
```
Arinono: REJECTED 🚮

PR #352, opened 2020-02-03
Chore/release 0.6.0

JeanBarriere:
This branch contains the dev state ready for the release v0.6.0


Arinono:
Not needed anymore as we want to sync the studio to the [storyscript-chart](https://github.com/storyscript/storyscript-chart), and we will need to skip `0.6.x` for that and jump directly to `0.7.1`

PR #378, opened 2020-02-08
feat(components): add Markdown type new chapters

Arinono:


Arinono:
https://github.com/storyscript/studio/pull/377 Goes first. Or we need to merge this one on `feat/add-chapter-story`, not `dev`

PR #386, opened 2020-02-11
feat(shade-clickable): add new class and new css to shade on hover

z0rzi:
This _should_ work fine, but now knowing Vue.js, I'm not sure if I did it the right way, especially concerning the position of the style (_should it be in a separate file?_)


src/app/views/Layout/Navbar.vue
```diff
@@ -17,7 +17,7 @@
         color="text-white"
         width="24"
         height="24"
-        class="mr-5"
+        class="mr-5 shade-on-hover"
```
Arinono:
```suggestion
        class="mr-5 hover:opacity-80"
```

src/app/views/Login.vue
```diff
@@ -1,5 +1,8 @@
 <template>
-  <div id="signup">
+  <div
+    id="signup"
+    class="shade-on-hover"
```
Arinono:
```suggestion
    class="hover:opacity-80"
```

src/internal/components/HelpDrop.vue
```diff
@@ -4,7 +4,7 @@
   >
     <div
       id="help"
-      class="flex items-center justify-center cursor-pointer"
+      class="flex items-center justify-center cursor-pointer shade-on-hover"
```
Arinono:
```suggestion
      class="flex items-center justify-center cursor-pointer hover:opacity-80"
```

PR #395, opened 2020-02-13
feat(footer): add footer under the editor pane with its new style

Arinono:
I removed the links for now since they lead to nothing (except the `Jobs` one)

Screenshot 2020-02-13 at 15 33 49


src/app/views/Editor/EditorPane.vue
```diff
@@ -3,7 +3,7 @@
     id="editor-pane"
     class="flex flex-1 flex-col"
   >
-    <div class="bg-white flex flex-col flex-1 rounded-t-md border-t border-l border-r border-gray-20">
+    <div class="min-h-screen bg-white flex flex-col flex-1 rounded-t-md border border-gray-20">
```
Arinono: Minimum height applied here so the footer is not visible unless we scroll passed this size.

>**I see** the Story takes up the full height of the window (regardless of the windows height)
**When** I scroll the footer is revealed according to the design

src/internal/components/Footer.vue
```diff
@@ -1,124 +1,88 @@
...
+      weight="medium"
+      class="text-gray-60 w-1/3"
+    >
+      © {{ new Date().getFullYear() }} Storyscript, Inc.
+    </s-text>
     <div
-      v-if="full"
-      class="flex flex-col sm:flex-row w-full max-w-5xl mx-auto pt-12 pb-6"
+      id="logo"
+      class="my-4 w-1/3 flex items-center justify-center"
```
Arinono: Fixed with on the 3 parts so they are well positioned when the links are not displayed. Could be removed once we have the links so the flex boxes can do their job alone.

src/internal/components/Footer.vue
```diff
@@ -1,124 +1,88 @@
...
+        fill="none"
+        xmlns="http://www.w3.org/2000/svg"
       >
-        <s-text
-          v-if="section.name"
-          class="mb-8 text-indigo-100"
-          p="1"
-          weight="semibold"
-          v-text="section.name"
+        <path
```
Arinono: I left the plain logo here for another story that will require moving some of its parts. And, for now, this behaviour will be only for the footer, so there's no need to move this SVG to another component/file.

PR #398, opened 2020-02-14
feat(footer): add magic wand

Arinono:


Arinono:
Not going to be finished here as it was like an exploration. Closing

PR #409, opened 2020-02-18
Bug/no sls in read only

z0rzi:


src/app/views/Editor/EditorPane.vue
```diff
@@ -98,7 +99,7 @@ import SFooter from '@internal/components/Footer.vue'
 })
 export default class EditorPane extends Vue {
   private editors: Array<Editor> = []
-  private storyscriptSLS: StoryscriptSLS = new StoryscriptSLS()
+  private storyscriptSLS: StoryscriptSLS | null = null
```
JeanBarriere: what you can also do here, is using `undefined` instead of null, so saying:
`private storyscriptSLS?: StoryscriptSLS` would make it undefined at first
and in your watcher you can set it to a `new StoryscriptSLS()` or to `undefined`

src/app/views/Editor/EditorPane.vue
```diff
@@ -199,18 +213,20 @@ export default class EditorPane extends Vue {
     }, 300)
   }
 
+  constructor () {
+    super()
+    if (!this.readonly) {
+      this.storyscriptSLS = new StoryscriptSLS()
+    }
+  }
+
```
JeanBarriere: use the Vue lifecycle methods here, whether in `mounted` or in `created`. Having the watcher on readonly being immediate should actually be sufficient in this case

src/app/views/Editor/EditorPane.vue
z0rzi: Oops, I kind of mixed up my git branches, my bad :no_mouth:

I created a new PR with the requested changes:
https://github.com/storyscript/studio/pull/420

z0rzi: Done in the last commit

src/app/views/Editor/EditorPane.vue
z0rzi: Done in the last commit

